PR #7, opened 2018-10-31
Added support for fetching QE Identity information

soccerGB:
This PR adds two new functions: sgx_get_qe_identity_info() and sgx_free_qe_identity_info() to support fetching Quote Enclave Identity Information. They will be called from Intel SGX DCAP library from Open Enclave SDK's oe_verify_report API implementation.

Before the Azure PCK service supports caching qe identity information, the changes in this PR fetch Quote Enclave Identity information directly from Intel server "https://api.trustedservices.intel.com/sgx/certification/v1/qe/identity".

### DIFF
--- a/src/dcap_provider.cpp
+++ b/src/dcap_provider.cpp
@@ -36,6 +36,9 @@ constexpr char PCK_CERT_ISSUER_CHAIN[] = "SGX-PCK-Certificate-Issuer-Chain";
 constexpr char CRL_ISSUER_CHAIN[] = "SGX-PCK-CRL-Issuer-Chain";
 constexpr char TCB_INFO_ISSUER_CHAIN[] = "SGX-TCB-Info-Issuer-Chain";
 constexpr char TCB_INFO[] = "SGX-TCBm";
+constexpr char CONTENT_TYPE[] = "Content-Type";
+constexpr char QE_ISSUER_CHAIN[] = "SGX-QE-Identity-Issuer-Chain";
+constexpr char REQUEST_ID[] = "Request-ID";
 };
 
 static constexpr char CYBERTRUST_ROOT_CERT[] =
@@ -67,10 +70,15 @@ static char DEFAULT_CERT_URL[] =
     "https://pck-cache-prod-webapp-eastus.azurewebsites.net/sgx/certificates";
 static std::string cert_base_url = DEFAULT_CERT_URL;
 
+// Before Azure PCK service supports caching qe identity, fetch directly from
+// Intel server for now
+static char QE_IDENTITY_URL[] =
+    "https://api.trustedservices.intel.com/sgx/certification/v1/qe/identity";
+
 #if 0 // Flip this to true for easy local debugging
 static void DefaultLogCallback(sgx_ql_log_level_t level, const char* message)
 {
-    printf("[%s]: %s\n", level == SGX_QL_LOG_ERROR ? "ERROR" : "DEBUG", message);
+    printf("Azure Quote Provider: libdcap_quoteprov.so [%s]: %s\n", level == SGX_QL_LOG_ERROR ? "ERROR" : "DEBUG", message);
 }
 
 static sgx_ql_logging_function_t logger_callback = DefaultLogCallback;
@@ -97,6 +105,53 @@ void log(sgx_ql_log_level_t level, const char* fmt, ...)
 
         logger_callback(level, message);
     }
+}
+
+//
+// get raw value for header_item item if exists
+//
+sgx_plat_error_t get_raw_header(
+    const curl_easy& curl,
+    const std::string header_item,
+    std::string* out_header)
+{
+    const std::string* raw_header = curl.get_header(header_item);
+    if (raw_header == nullptr)
+    {
+        log(SGX_QL_LOG_ERROR, "Header '%s' is missing.", header_item);
+        return SGX_PLAT_ERROR_UNEXPECTED_SERVER_RESPONSE;
+    }
+    if (out_header != nullptr)
+    {
+        *out_header = *raw_header;
+        log(SGX_QL_LOG_INFO,
+            "raw_header %s:[%s]\n",
+            header_item,
+            raw_header->c_str());
+    }
+    return SGX_PLAT_ERROR_OK;
+}
+
+//
+// get unescape value for header_item item if exists
+//
+sgx_plat_error_t get_unescape_header(
+    const curl_easy& curl,
+    const std::string header_item,
+    std::string* unescape_header)
+{
+    sgx_plat_error_t result = SGX_PLAT_ERROR_OK;
+    std::string raw_header;
+
+    result = get_raw_header(curl, header_item, &raw_header);
+    if (result != SGX_PLAT_ERROR_OK)
+        return result;
+    *unescape_header = curl.unescape(raw_header);
+    log(SGX_QL_LOG_INFO,
+        "unescape_header %s:[%s]\n",
+        header_item,
+        unescape_header->c_str());
+    return result;
 }
 
 //
@@ -243,6 +298,7 @@ static std::string build_cert_chain(const curl_easy& curl)
     const std::string chain =
         curl.unescape(*curl.get_header(headers::PCK_CERT_ISSUER_CHAIN));
 
+    log(SGX_QL_LOG_INFO, "libquote_provider.so: [%s]\n", chain.c_str());
     return leaf_cert + chain;
 }
 
@@ -292,12 +348,12 @@ static sgx_plat_error_t parse_svn_values(
     const curl_easy& curl,
     sgx_ql_config_t* quote_config)
 {
-    const std::string* tcb = curl.get_header(headers::TCB_INFO);
-    if (tcb == nullptr)
-    {
-        log(SGX_QL_LOG_ERROR, "TCB info header is missing.");
-        return SGX_PLAT_ERROR_UNEXPECTED_SERVER_RESPONSE;
-    }
+    sgx_plat_error_t result = SGX_PLAT_ERROR_OK;
+
+    std::string tcb;
+    result = get_raw_header(curl, headers::TCB_INFO, &tcb);
+    if (result != SGX_PLAT_ERROR_OK)
+        return result;
 
     // string size == byte size * 2 (for hex-encoding)
     static constexpr size_t CPUSVN_SIZE =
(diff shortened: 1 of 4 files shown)


src/dcap_provider.cpp
```diff
@@ -714,3 +721,112 @@ extern "C" sgx_plat_error_t sgx_ql_set_logging_function(
     logger_callback = logger;
     return SGX_PLAT_ERROR_OK;
 }
+
+extern "C" sgx_plat_error_t sgx_get_qe_identity_info(
+    sgx_qe_identity_info_t** pp_qe_identity_info)
+{
+    std::string intel_cert_base_url = INTEL_QE_IDENTITY_URL;
```
sethmoo: There's no need to create this string. I'd just pass `QE_IDENTITY_URL` everywhere.

src/dcap_provider.cpp
```diff
@@ -67,10 +70,13 @@ static char DEFAULT_CERT_URL[] =
     "https://pck-cache-prod-webapp-eastus.azurewebsites.net/sgx/certificates";
 static std::string cert_base_url = DEFAULT_CERT_URL;
 
+// Before Azure PCK service supports caching qe identity, fetch directly from Intel server
+static char INTEL_QE_IDENTITY_URL[] = "https://api.trustedservices.intel.com/sgx/certification/v1/qe/identity";
```
sethmoo: Please rename to `QE_IDENTITY_URL` because we'll want to swap this over to the Azure cache once we have support for it.

src/dcap_provider.cpp
```diff
@@ -714,3 +721,112 @@ extern "C" sgx_plat_error_t sgx_ql_set_logging_function(
...
+        return SGX_PLAT_ERROR_INVALID_PARAMETER;
+    }
+
+    try 
+    {
+        std::vector<uint8_t> identity_info;
+        std::string issuer_chain;
+        size_t total_buffer_size = 0;
+
+        const auto curl = curl_easy::create(intel_cert_base_url, CYBERTRUST_ROOT_CERT);
```
sethmoo: Intel does not use the cybertrust root. They seem to be using COMODO.

If this is working, then we have a bug with our TLS cert processing. Does this work? If so, let's open a bug and figure out why :)

src/dcap_provider.cpp
```diff
@@ -714,3 +721,112 @@ extern "C" sgx_plat_error_t sgx_ql_set_logging_function(
...
+        log(SGX_QL_LOG_INFO, "Fetching QE Identity from remote server: '%s'.", intel_cert_base_url.c_str());
+        curl->perform();
+
+        // content type
```
sethmoo: Why bother with getting content type? It's not used for anything.

src/dcap_provider.cpp
```diff
@@ -714,3 +721,112 @@ extern "C" sgx_plat_error_t sgx_ql_set_logging_function(
...
+            log(SGX_QL_LOG_ERROR,
+                "Header '%s' is missing.",
+                headers::QE_ISSUER_CHAIN);
+            return SGX_PLAT_ERROR_UNEXPECTED_SERVER_RESPONSE;
+        }
+        issuer_chain = curl->unescape(*issuer_chain_raw_header);
+        log(SGX_QL_LOG_INFO, "QE_ISSUER_CHAIN: [%s]\n", issuer_chain.c_str());
+
+        // requester id
+        std::string request_id = curl->unescape(*curl->get_header(headers::REQUEST_ID));
```
sethmoo: Missing null check (like above).

I think there should be a helper function to get unescaped headers that can catch missing headers, log them, and return `SGX_PLAT_ERROR_UNEXPECTED_SERVER_RESPONSE`.

src/dcap_provider.cpp
```diff
@@ -714,3 +721,112 @@ extern "C" sgx_plat_error_t sgx_ql_set_logging_function(
...
+        log(SGX_QL_LOG_INFO, "REQUEST_ID:[%s]\n", request_id.c_str());
+
+        // read body
+        identity_info = curl->get_body();
+        std::string qe_identity(curl->get_body().begin(), curl->get_body().end());
```
sethmoo: Let's remove this log for a couple of reasons:
1. This is making a full, in-memory copy of the buffer just to log it.
2. The log message is going to be extremely long. If the receiver of the buffer really wants to log it, they are free to do so themselves.

src/dcap_provider.cpp
```diff
@@ -714,3 +721,112 @@ extern "C" sgx_plat_error_t sgx_ql_set_logging_function(
...
+        if (!buffer)
+            return SGX_PLAT_ERROR_OUT_OF_MEMORY;
+
+        memset(buffer, 0, total_buffer_size);
+
+        // fill in the qe info
+        p_qe_identity_info = reinterpret_cast<sgx_qe_identity_info_t*>(buffer);
+
+        // advance to the end of the sgx_qe_identity_info_t structure
+        buffer += sizeof(*p_qe_identity_info);
```
sethmoo: Please add assertions to ensure you're not walking past the end of your buffer. See `sgx_ql_get_revocation_info` for a sample.

src/dcap_provider.cpp
```diff
@@ -714,3 +721,112 @@ extern "C" sgx_plat_error_t sgx_ql_set_logging_function(
...
+        std::string qe_identity(curl->get_body().begin(), curl->get_body().end());
+        log(SGX_QL_LOG_INFO, "body:[%s]\n", qe_identity.c_str());
+
+        // Calculate total buffer size
+        total_buffer_size = safe_add(sizeof(sgx_qe_identity_info_t), identity_info.size());
+        total_buffer_size = safe_add(total_buffer_size, 1); // null terminator
+        total_buffer_size = safe_add(total_buffer_size, issuer_chain.size());
+
+        buffer = new char[total_buffer_size];
+        if (!buffer)
```
sethmoo: FYI - This version of `new` will never return `nullptr`; it will throw `std::bad_alloc`.

We've already got a `bad_alloc` exception handler at the end of the function. Let's just rely on that to catch the OOM.

src/dcap_provider.cpp
```diff
@@ -714,3 +721,112 @@ extern "C" sgx_plat_error_t sgx_ql_set_logging_function(
...
+        p_qe_identity_info->qe_id_info = buffer;
+        memcpy(p_qe_identity_info->qe_id_info,
+                identity_info.data(),
+                identity_info.size());
+        buffer += identity_info.size() + 1; // skip null terminator
+
+        // set issuer_chain info
+        p_qe_identity_info->issuer_chain_size =  static_cast<uint32_t>(issuer_chain.size());
+        p_qe_identity_info->issuer_chain = buffer;
+        buffer += issuer_chain.size() + 1; // skip null terminator
```
sethmoo: This null terminator is not accounted for above in the buffer size arithmetic.

IMO, this would be a great use case for assertions, because you would have gotten an abort if the function ended with `assert(buffer == buffer_end);` :)

src/dcap_provider.cpp
```diff
@@ -714,3 +721,112 @@ extern "C" sgx_plat_error_t sgx_ql_set_logging_function(
...
+        *pp_qe_identity_info = p_qe_identity_info;
+    }
+    catch (std::bad_alloc&)
+    {
+        return SGX_PLAT_ERROR_OUT_OF_MEMORY;
+    }
+    catch (std::overflow_error& error)
+    {
+        log(SGX_QL_LOG_ERROR, "Overflow error. '%s'", error.what());
+        delete[] buffer;
```
sethmoo: I'm pretty sure buffer will always be null at this point. However, it doesn't hurt to catch and clean up.

That said, if we're going to clean up in one handler, I think we should clean up in _all_ of them.

src/dcap_provider.h
```diff
@@ -68,6 +68,24 @@ typedef sgx_plat_error_t (*sgx_ql_get_revocation_info_t)(
 typedef void (*sgx_ql_free_revocation_info_t)(
     sgx_ql_revocation_info_t* p_revocation_info);
 
+/*****************************************************************************
+ * Data types and interfaces for getting qe identity info 
+ ****************************************************************************/
+
+typedef struct _sgx_qe_identity_info_t
```
sethmoo: Please document that the buffers are null-terminated.

src/dcap_provider.cpp
soccerGB: good catch. I will update it.

src/dcap_provider.cpp
soccerGB: Thanks for the information, I will update accordingly

src/dcap_provider.h
soccerGB: good point.

soccerGB: added

src/dcap_provider.cpp
soccerGB: An 2nd look at it, I agree that buffer will be null in the "catch std::overflow_error" block because all the safe_add operations happened above the memory allocation for buffer. I will remove the delete.

src/dcap_provider.cpp
soccerGB: advice taken

src/dcap_provider.cpp
soccerGB: addressed

src/dcap_provider.cpp
soccerGB: addressed

src/dcap_provider.cpp
soccerGB: “Retrieve QE Identity” does not require any client authentication.

src/dcap_provider.cpp
soccerGB: It was there for debugging purpose. I will remove it.

src/dcap_provider.cpp
soccerGB: Checking added

src/dcap_provider.cpp
soccerGB: Agree. It was for debugging purpose. I will remove it.

src/dcap_provider.cpp
sethmoo: This is server auth, not client auth.

src/dcap_provider.cpp
sethmoo: Please add a helper function instead of duplicating code.

I know it's only ~10 loc, but any form of duplication is bad news.

src/dcap_provider.h
sethmoo: Thanks for adding the docs above. Please be sure to also indicate `qe_id_info` and `issuer_chain` below are null terminated.

soccerGB: added

src/dcap_provider.cpp
soccerGB: Okay! :-)
Added get_unescape_header() and get_raw_header().

src/dcap_provider.cpp
soccerGB: Replaced with the cert you provided.

soccerGB: Undid the added cert.

soccerGB: Please see my comments on the following issue:
https://github.com/Microsoft/Azure-DCAP-Client/issues/8

sethmoo: We still need the correct certificate, even if issue #8 remains open.

Issue #8, opened 2018-11-01
Root cert validation is not working correctly


sethmoo:
The root cert, as passed to `curl_easy::create` is not being checked as a valid root for https connections.

It appears we have a bug in `ssl_context_callback`. Address this and add a negative test case to ensure we don't run across this again in the future.

soccerGB:
I took a look at the curl_easy class implementation. It looks like the ca_certparameter passed in the call to curl_easy::create(intel_cert_base_url, ca_cert) was used as the client's cert, which was handled in ssl_context_callback.

To validate server's certificate, 

// verify the server's SSL certificate, we can add the following two CURL options in the easy_curl implementation to enable checking on the server certificate and host name

    // verify the server's SSL certificate
    easy->set_opt_or_throw(CURLOPT_SSL_VERIFYPEER, 1L);
    // verify server certificate's name against host
    easy->set_opt_or_throw(CURLOPT_SSL_VERIFYHOST, 2L);

sethmoo:
You also need the code in `ssl_context_callback` to setup the root cert store. It just looks like the correct SSL verify options weren't also set.

sethmoo:
It turns out we didn't have a problem here. There is a default root cert store for libcurl, and the necessary roots are included.

PR #10, opened 2018-11-08
Remove ca_cert from curl_easy::create

sethmoo:
We don't need to pass an explicit root, as the default root certificate
store used by curl has the correct roots. This allows us to remove not
only the explicit root from our code, but the SSL_CTX_* curl options
as well.

Curl also does peer and hostname validation by default, so we don't
need to explicitly set those parameters.


sethmoo:
This addresses issue #8 

It turns out root cert validation wasn't working as intended, but it doesn't matter. We can just rely on the default cert store.

soccerGB:
Change look good and it passed all the OE test cases

PR #17, opened 2018-11-14
Properly use custom libdcap_quoteprov.so build for CI

ionutbalutoiu:
In order to make sure that the custom `libdcap_quoteprov.so` library is used by the CI, we have to set `LD_PRELOAD` environment variable as well.


soccerGB:
The LD_LIBRARY_PATH change in the current master branch solves this CI specific issue.
We are good without changes in this PR.

PR #27, opened 2019-01-08
Add dynamic user and group to docker images

stefanlupsa:
The idea is to dynamically get the Jenkins UID GID user name and group name to create a valid user inside the image.

The Dockerfile was hard-coded with Jenkins user and group (1001) and that may not be the case, as well in the Jenkinsfile the user / group was not passed when the images are built and if the Jenkins user is not 1001 this would create the wrong user inside the docker image.


ionutbalutoiu:
CI is green. This is ready to be merged.

PR #28, opened 2019-01-16
Make jenkins use credentials for oetools docker image

stefanlupsa:
Calling docker OETOOLS_IMAGE must be enclosed with docker.withRegistry block.
Otherwise, if the private OETOOLS_IMAGE is not pre-pulled, the job will fail.


ionutbalutoiu:
LGTM

PR #35, opened 2019-01-31
Published a version 1.1 of the Azure DCAP Client library

msft-jonnadul:


src/Linux/debian/changelog
```diff
@@ -3,3 +3,11 @@ az-dcap-client (1.0) unstable; urgency=medium
   * Public release
 
  -- "Seth Moore" <sethmo@microsoft.com>  Mon, 28 Sep 2018 09:00:00 -0700
+
```
soccerGB: The log for the new version (1.1) need to be moved to above 1.0 to avoid warning like the following during the  dpkg-buildpackage operation


dpkg-genchanges: warning: the current version (1.0) is earlier than the previous one (1.1)

PR #41, opened 2019-02-13
Remove root user access in container when running on host folder

stefanlupsa:
Besides the fact that it's not OK to use the `-u root` option for all commands, some runs would leave artifacts in the Jenkins work space even with line used previously to remove the build folder.

This change moves the installation steps of the image into an intermediary `Dockerfile.scripts` which must be run without docker cache on top of the oetools-tools image that is cached locally in order to install the currently built .deb package of az-dcap-client (and the open-enclave release, since it only unpacks in the /opt folder can use the same `Dockerfile.scripts` for both stages) and then run the tests in a container from that image with the Jenkins user in the mounted work-space folder

### DIFF
--- a/.jenkins/Dockerfile
+++ b/.jenkins/Dockerfile
@@ -1,3 +1,5 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
 ARG UBUNTU_VERSION=16.04
 FROM ubuntu:${UBUNTU_VERSION}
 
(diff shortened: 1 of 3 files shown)


stefanlupsa:
here: 
https://oe-jenkins.eastus.cloudapp.azure.com/blue/organizations/jenkins/Az-DCAP/detail/PR-41/17/pipeline is a run with jenkins user added inside the Dockerfile.scripts 
and here: 
https://oe-jenkins.eastus.cloudapp.azure.com/blue/organizations/jenkins/Az-DCAP/detail/PR-41/19/pipeline 
is a run with extra env var AZDCAPCACHE added to Jenkinsfile

Both solutions fix the CI side of the problem when removing root user in the container but the other side should be handled in a different PR, i would rather use the env var as it's more elegant but it's not really documented

PR #44, opened 2019-02-15
Add unified docker build-args to all stages

stefanlupsa:
Since the docker images need about the same build args we can add them only once in order to make the code more readable,

Add groovy function to concatenate all docker build args in a more organized manner.

Unified the Dockerfile `ubuntu_version` to be without caps like in the open-enclave image Dockerfile so we can build with the same build args the Jenkins user args should not be consumed by the open-enclave Dockerfile since they are not defined in it


ionutbalutoiu:
LGTM

PR #47, opened 2019-02-25
New environment variable to specify clientid

msft-jonnadul:


src/dcap_provider.cpp
```diff
@@ -23,7 +23,9 @@
 #include <winsock.h>
 #endif
 
-#define MAX_URL_LENGTH 2000
+#define ENV_AZDCAP_BASE_URL "AZDCAP_BASE_CERT_URL"
+#define ENV_AZDCAP_CLIENT_ID "AZDCAP_CLIENT_ID"
```
soccerGB: I am wondering the impact difference between those two environment vars.

src/dcap_provider.cpp
```diff
@@ -88,24 +90,50 @@ void log(sgx_ql_log_level_t level, const char* fmt, ...)
...
 
+static std::string get_client_id()
+{
+    const char * env_client_id = getenv(ENV_AZDCAP_CLIENT_ID);
+    if (env_client_id == NULL)
+    {
+        return "";
+    }
+    
+    if ((env_client_id != NULL) && (strnlen(env_client_id, MAX_ENV_VAR_LENGTH) == MAX_ENV_VAR_LENGTH))
```
soccerGB: The checking for "(env_client_id != NULL)" seems to be redundant.

src/dcap_provider.cpp
```diff
@@ -300,7 +328,7 @@ static std::string build_pck_cert_url(const sgx_ql_pck_cert_id_t& pck_cert_id)
         format_as_big_endian_hex_string(pck_cert_id.pce_id);
 
     return get_base_url() + '/' + qe_id + '/' + cpu_svn + '/' + pce_svn + '/' +
-           pce_id + '?' + API_VERSION;
+           pce_id + '?' + "clientid=" + get_client_id() + '&' + API_VERSION;
 }
```
soccerGB: It looks like the change make "clientid=" a required parameter for PCK Caching service, right?
That is, it needs to treat pce_svn/pce_id/clientid=/API_VERSION a valid request string.

johnkord: Yes this is a good point. Is this field going to be required in the future? Maybe, if the environment variable isn't set, we shouldn't even reference the "clientid" field in the URL?  What do you think would be best @msft-jonnadul?

msft-jonnadul: I confirmed that append a undefined clientid value to the calls into the PCK cache end point don't seem to regress previous behavior, ex: this query will resolve currently https://global.acccache.azure.net/sgx/certificates/00bd4b2879d5a2764a964ab990908b67/05050205ff8001000000000000000000/0600/0000?clientid=&api-version=2018-10-01-preview.

msft-jonnadul: But I'll modify it such that if a clientid isn't defined, then it'll not even attempt to populate this variable.

src/dcap_provider.cpp
msft-jonnadul: Can you elaborate? If only specified to unique set of known client id values, shouldn't affect anything else in the PCK Cache server's behavior drastically.

soccerGB: I am sorry for not being more clear on my original question. I meant to ask what would end up being tested when different values are set in those two environment variables. I would be great to add comments in the code. Also, adding some log messages to give some hit on the values in use.

msft-jonnadul: Agreed with this. I added comments above the environment variable definitions outlining the intent of these values, and get_base_url() and get_client_id() have logging in-place indicating whether values from these environment variables have been set and being consumed.

soccerGB: Thanks for addressing my comments.

PR #48, opened 2019-04-05
Modified dcap client to fetch qeid from pck cache service, and modifi…

msft-jonnadul:
…ed some logging statuses


msft-jonnadul:
> Changes look good.
> It would be great to see signature validation tests added.

Yes, we'll plan to add that on the server side.

PR #59, opened 2019-05-07
Build fixes; got working on WIndows

LarryOsterman:


src/curl_easy.cpp
```diff
@@ -84,6 +88,27 @@ std::unique_ptr<curl_easy> curl_easy::create(const std::string& url)
...
+    char fileNameBuffer[MAX_PATH];
+    if (GetModuleFileNameA(moduleHandle, fileNameBuffer, _countof(fileNameBuffer)) == 0)
+    {
+        throw std::exception("Unable to retrieve module name.");
+    }
+    if (!PathRemoveFileSpecA(fileNameBuffer))
+    {
+        throw std::exception("Unable to remove filename from buffer.");
+    }
+    if (FAILED(StringCchCatA(fileNameBuffer, _countof(fileNameBuffer), "\\curl-ca-bundle.crt")))
```
soccerGB: I am wondering why a fix file name is used here.
Is this for testing purpose?

anitagov: Yes, this is the file from copied from libcurl.

anitagov:
Adding @CodeMonkeyLeet ...

src/curl_easy.cpp
LarryOsterman: LibCURL requires this filename.

PR #71, opened 2019-06-07
Setting a production default client id

msft-jonnadul:


msft-jonnadul:
@anitagov , agreed that we should refactor it but will clean that up in a separate pull request.

PR #73, opened 2019-06-12
Unit testing for windows dcap

msft-chow:


src/UnitTests/test_quote_prov.cpp
```diff
@@ -1,25 +1,34 @@
...
-#include "UnitTests/unit_test.h"
 
+#if defined(__LINUX__)
+#include <dlfcn.h>
+#include <tgmath.h>
 #include <cstdio>
 #include <cstring>
 #include <ctime>
 #include <memory>
-#include <dlfcn.h>
```
LarryOsterman: Why are you reordering these headers?

src/UnitTests/test_quote_prov.cpp
```diff
@@ -1,25 +1,34 @@
...
-#include <tgmath.h>
-
-#include "intel/sgx_ql_lib_common.h"
+#else
+#include <iostream>
+#include <stdio.h>
```
LarryOsterman: This is C++ code, why are you using the C runtime library functions instead of their C++ equivalents (stdio.h vs cstdio, string.h vs cstring, etc)? That allows you to have significant commonality with the Linux header block.

src/UnitTests/test_quote_prov.cpp
```diff
@@ -1,25 +1,34 @@
...
+#include <time.h>
+#include <windows.h>
+#include <memory>
+#endif
 
 typedef quote3_error_t (*sgx_ql_get_quote_config_t)(
-    const sgx_ql_pck_cert_id_t *p_pck_cert_id,
-    sgx_ql_config_t **pp_quote_config);
+    const sgx_ql_pck_cert_id_t* p_pck_cert_id,
+    sgx_ql_config_t** pp_quote_config);
```
LarryOsterman: Have you consciously chosen to align this code with the C++ coding conventions (http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#main) or is this a gratutous change throughout the file? If this was made to align with C++ coding conventions, that's cool, but if so, you should probably do that in a separate commit.

src/UnitTests/test_quote_prov.cpp
```diff
@@ -33,50 +42,99 @@ static sgx_ql_set_logging_function_t sgx_ql_set_logging_function;
...
+        reinterpret_cast<sgx_ql_set_logging_function_t>(
+            dlsym(library, "sgx_ql_set_logging_function"));
     assert(sgx_ql_set_logging_function);
 
     return library;
 }
+#else
+static HINSTANCE LoadFunctions()
+{
+    // https://stackoverflow.com/questions/27220/how-to-convert-stdstring-to-lpcwstr-in-c-unicode
```
msft-jonnadul: Nit: remove

src/UnitTests/test_quote_prov.cpp
```diff
@@ -33,50 +42,99 @@ static sgx_ql_set_logging_function_t sgx_ql_set_logging_function;
...
+    std::wstring s = L"dcap_quoteprov.dll";
+    HINSTANCE hLibCapdll = LoadLibrary(s.c_str());
+    if (hLibCapdll == NULL)
+    {
+        DWORD error = GetLastError();
+        std::cout << error << std::endl;
```
msft-jonnadul: remove all references to std::cout and replace with Log(..).

src/UnitTests/test_quote_prov.cpp
```diff
@@ -33,50 +42,99 @@ static sgx_ql_set_logging_function_t sgx_ql_set_logging_function;
...
-        levelText,
-        message);
+    printf("[%s]: %s\n", levelText, message);
 }
 
+#if defined __LINUX__
 static void* LoadFunctions()
 {
-    // if this assert fails, be sure libdcap_quoteprov.so is in your LD_LIBRARY_PATH
+    // if this assert fails, be sure libdcap_quoteprov.so is in your
```
LarryOsterman: Why reindent the line?

src/UnitTests/test_quote_prov.cpp
```diff
@@ -33,50 +42,99 @@ static sgx_ql_set_logging_function_t sgx_ql_set_logging_function;
...
     void* library = dlopen("libdcap_quoteprov.so", RTLD_NOW);
+    printf(dlerror());
     if (library == nullptr)
     {
         Log(SGX_QL_LOG_ERROR, dlerror());
         abort();
     }
 
-    sgx_ql_free_revocation_info = reinterpret_cast<sgx_ql_free_revocation_info_t>(dlsym(library, "sgx_ql_free_revocation_info"));
+    sgx_ql_free_revocation_info =
```
LarryOsterman: The parenthesising on this is somewhat unusual.

src/UnitTests/test_quote_prov.cpp
```diff
@@ -33,50 +42,99 @@ static sgx_ql_set_logging_function_t sgx_ql_set_logging_function;
...
+        reinterpret_cast<sgx_ql_free_revocation_info_t>(
+            dlsym(library, "sgx_ql_free_revocation_info"));
     assert(sgx_ql_free_revocation_info);
 
-    sgx_ql_get_revocation_info = reinterpret_cast<sgx_ql_get_revocation_info_t>(dlsym(library, "sgx_ql_get_revocation_info"));
+    sgx_ql_get_revocation_info = reinterpret_cast<sgx_ql_get_revocation_info_t>(
+        dlsym(library, "sgx_ql_get_revocation_info"));
```
LarryOsterman: Why not isolate the GetProcAddress/dlsym functions into a common function which implements the functionality - that way the Linux and Windows code which does the loading from the shared object can be the same.

src/UnitTests/test_quote_prov.cpp
```diff
@@ -33,50 +42,99 @@ static sgx_ql_set_logging_function_t sgx_ql_set_logging_function;
...
+    // https://stackoverflow.com/questions/27220/how-to-convert-stdstring-to-lpcwstr-in-c-unicode
+    std::wstring s = L"dcap_quoteprov.dll";
+    HINSTANCE hLibCapdll = LoadLibrary(s.c_str());
+    if (hLibCapdll == NULL)
+    {
+        DWORD error = GetLastError();
+        std::cout << error << std::endl;
+
+        //Log(SGX_QL_LOG_ERROR, error);
+        exit(0);
```
LarryOsterman: Why not call abort() here?

src/UnitTests/test_quote_prov.cpp
```diff
@@ -129,19 +225,20 @@ static void GetCrlTest()
...
 
-    static constexpr uint8_t TEST_FMSPC[] = { 0x00, 0x90, 0x6E, 0xA1, 0x00, 0x00 };
+    static constexpr uint8_t TEST_FMSPC[] = {
+        0x00, 0x90, 0x6E, 0xA1, 0x00, 0x00};
 
     // This is the CRL DP used by Intel for leaf certs
-    static const char* TEST_CRL_URL = "https://api.trustedservices.intel.com/sgx/certification/v1/pckcrl?ca=processor";
+    static const char* TEST_CRL_URL =
+        "https://api.trustedservices.intel.com/sgx/certification/v1/"
+        "pckcrl?ca=processor";
```
LarryOsterman: Why is this split across multiple lines?

src/UnitTests/test_quote_prov.cpp
msft-chow: Careless mistake on my part. noted and reverting it

src/UnitTests/test_quote_prov.cpp
msft-chow: I don't remember consciously doing this. I believe lots of these are also done by visual studio when they automatically pretty up code

src/Windows/UnitTest/dcap_provider_ut_example.sln
```diff
@@ -0,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+VisualStudioVersion = 15.0.28307.572
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "dcap_provider_ut_example", "dcap_provider_ut_example.vcxproj", "{C356FCD3-13E4-40F1-9FDD-D2E68FA97885}"
```
msft-jonnadul: Nit: Rename dcap_provider_ut_example -> dcap_provider_ut

src/UnitTests/test_quote_prov.cpp
msft-chow: reverted

src/UnitTests/test_quote_prov.cpp
msft-chow: done

src/Windows/UnitTest/dcap_provider_ut_example.sln
msft-chow: renamed all references

src/UnitTests/test_quote_prov.cpp
msft-chow: reverted

src/UnitTests/test_quote_prov.cpp
msft-chow: done

.gitignore
```diff
@@ -3,6 +3,9 @@
 ##
 ## Get latest from https://github.com/github/gitignore/blob/master/VisualStudio.gitignore
 
+src/Linux/ext/intel/*
+*.o
```
LarryOsterman: Why only exclude .o files in src/Linux/ext/intel and not all .o files (similar to the blanket exemption to .obj files below)? And why only .o files - there are other intermediate files (a.out for instance) on linux that should probably be ignored.

.jenkins/Jenkinsfile
```diff
@@ -107,11 +107,13 @@ def DCAPBuildTest(String label, String build_type) {
                 checkout scm
                 bat """
                     vcvars64.bat x64 && \
-                    cd ${WORKSPACE}\\src\\Windows && \
-                    powershell.exe -ExecutionPolicy Bypass -Command \".\\get-prereqs.ps1\" && \
```
LarryOsterman: Why don't you still need get-prereqs.ps1?

src/UnitTests/test_quote_prov.cpp
```diff
@@ -77,6 +83,36 @@ static void* LoadFunctions()
...
+    sgx_ql_free_quote_config = reinterpret_cast<sgx_ql_free_quote_config_t>(GetProcAddress(hLibCapdll, "sgx_ql_free_quote_config"));
+    assert(sgx_ql_free_quote_config);
+
+    sgx_ql_get_quote_config = reinterpret_cast<sgx_ql_get_quote_config_t>(GetProcAddress(hLibCapdll, "sgx_ql_get_quote_config"));
+    assert(sgx_ql_get_quote_config);
+
+    sgx_ql_set_logging_function = reinterpret_cast<sgx_ql_set_logging_function_t>(GetProcAddress(hLibCapdll, "sgx_ql_set_logging_function"));
+    assert(sgx_ql_set_logging_function);
+
+    return hLibCapdll;
```
soccerGB: Do you need to call FreeLibrary here?

src/UnitTests/test_quote_prov.cpp
```diff
@@ -33,26 +39,26 @@ static sgx_ql_set_logging_function_t sgx_ql_set_logging_function;
...
         case SGX_QL_LOG_WARNING:
             levelText = "WARNING";
             break;
         case SGX_QL_LOG_INFO:
             levelText = "INFO";
             break;
         default:
-	    levelText = "ERROR";
+            levelText = "ERROR";
     }
```
soccerGB: A break in this default case should be good enough because the default value of levelText was initialized to ERROR.

src/UnitTests/test_quote_prov.cpp
```diff
@@ -158,6 +192,7 @@ static void GetCrlTest()
     TEST_PASSED();
 }
 
+#if defined __LINUX__
 extern void QuoteProvTests()
```
soccerGB: I am curious on why this was excluded from Windows tests

src/UnitTests/test_quote_prov.cpp
```diff
@@ -111,11 +146,12 @@ static void GetCertsTest()
         0x05, 0x05, 0x02, 0x04, 0x01, 0x80, 0x01, 0x00,
         0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00
     };
+
```
soccerGB: This comment is not specific to Windows test case. It applies to both linux and Windows.

This test seems to assume the specific qe_id to have specific CPUSVN values, which are not fixed values. This test will fail when the CPUSVN is updated.

src/UnitTests/test_quote_prov.cpp
```diff
@@ -184,14 +220,32 @@ extern void QuoteProvTests()
...
-    // Ensure that there is a signficiant enough difference between the cert fetch
-    // to the end point and cert fetch to local cache and that local cache call is
-    // fast enough
-    assert (fabs(duration_curl - duration_local) > 0.0001 && duration_local < 0.001);
-    
+    // Ensure that there is a signficiant enough difference between the cert
+    // fetch to the end point and cert fetch to local cache and that local cache
+    // call is fast enough
+    assert(fabs(duration_curl - duration_local) > 0.0001 && duration_local < 0.001);
+
```
soccerGB: I am wondering why those 0.0001 and 0.001 were pick.
The timing test like this is usually not stable

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,33 @@
...
-    return nullptr;
-}
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "local_cache.h"
+
+//
+// The Windows implementation of the local cache is not yet complete.
+// All methods noop, so we'll never cache any data locally.
```
soccerGB: It would be great to file an issue to track this pending work for implementing local caching on Windows.

src/UnitTests/test_quote_prov.cpp
msft-jonnadul: This was necessary because we don't have full support for local caching in place for the Azure DCAP for windows yet. @msft-chow will be working on that next.

src/UnitTests/test_quote_prov.cpp
msft-jonnadul: As long as we don't remove this entry from the production PCK Caching Service, this should still work right?

src/Windows/dll/local_cache.cpp
msft-jonnadul: Issue created, https://github.com/microsoft/Azure-DCAP-Client/issues/75.

src/UnitTests/test_quote_prov.cpp
msft-jonnadul: Yea this was my initial attempt at capturing a perf distinction between fetches from the PCK Caching Service and the Local cache. I'm open to other approach for testing out this particular scenario.

src/UnitTests/test_quote_prov.cpp
soccerGB: is that qe_id for a real machine? Or a dummy one?
If it's a dummy, it should be ok as long as you keep it in your table.

msft-jonnadul: Its a dummy one I believe.

src/UnitTests/test_quote_prov.cpp
soccerGB: If the purpose is to make sure the data is actually fetched from local cache, is it possible to do it via checking the delta of the cache contents before  and after test operations instead of the fetched time?

src/Windows/dll/local_cache.cpp
soccerGB: Thanks for addressing this.

src/UnitTests/test_quote_prov.cpp
msft-jonnadul: Ok we'll plan to address that as part of @msft-chow's upcoming pull request for local cache related changes.

PR #76, opened 2019-07-16
Local Cache for Windows

msft-chow:


src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+    {
+        if ((buf.st_mode & _S_IFDIR) > 0)
+        {
+            return;
+        }
+
+        throw std::runtime_error(dirname + " already exists, and is not a directory.");
+    }
+
+    wchar_t dirName[64];
```
LarryOsterman: Instead of 64, use MAX_PATH.

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+    std::mbstowcs(dirName, dirname.c_str(), strlen(dirname.c_str()) + 1);//Plus null
+
+    rc = CreateDirectory(dirName, NULL);
+    if (rc != 0)
```
LarryOsterman: Use GetLastError() to provide the actual error code for callers.

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+    {
+        throw_errno("Error creating directory '" + dirname + "'");
```
LarryOsterman: throw_errno assumes that you're only ever calling C runtime library functions. Since you used CreateDirectory, errno won't be set with the correct error code.

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+    }
+}
+
+static void init_callback()
+{
+    const char * env_home = ::getenv("APPDATA");
```
LarryOsterman: Is appdata the right directory to use? That's the roaming state directory, shouldn't you use LOCALAPPDATA instead?

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+    }
+    else if (env_home != 0 && (strcmp(env_home, "") != 0))
+    {
+        dirname = std::string(env_home);
+    }
+    else
+    {
+        // Throwing exception if the expected HOME
+        // environment variable is not defined.
+        throw std::runtime_error("HOME and AZDCAPCACHE environment variables not defined");
```
LarryOsterman: This says it depends on an environment variable "HOME", but instead it depends on APPDATA.

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+    }
+
+    dirname += application_name;
+    make_dir(dirname);
```
LarryOsterman: can make_dir ever fail? If so, check for failure.

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+        (PBYTE)&cbHashObject,
+        sizeof(DWORD),
+        &cbData,
+        0)))
+    {
+        errorString = "Error 0x" + std::to_string(status) + "returned by BCryptGetProperty\n";
+        goto Cleanup;
+    }
+
+    pbHashObject = (PBYTE)HeapAlloc(GetProcessHeap(), 0, cbHashObject);
```
LarryOsterman: Why not use std::unique_ptr<BYTE> here? That means you don't have to manually free the buffer.

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+        pbHash,
+        cbHash,
+        0)))
+    {
+        errorString = "Error 0x" + std::to_string(status) + "returned by BCryptFinishHash\n";
+        goto Cleanup;
+    }
+
+    retval.reserve(2 * cbHash + 1);
+    for (size_t i = 0; i < cbHash; i++)
```
LarryOsterman: Why use .reserve? Why not just rely on the fact that the std::string::push_back grows the string.

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+    {
+        char buf[3];
+        snprintf(buf, sizeof(buf), "%02x", pbHash[i]);
```
LarryOsterman: Why use sprintf here? Why not manually create the string?

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+{
+    return sha256(input.length(), input.data());
+}
+
+static std::string get_file_name(const std::string& id)
+{
+    return g_cache_dirname + "\\" + sha256(id);
+}
+
+void local_cache_clear() {
```
LarryOsterman: Use a consistent brace style.

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+    std::wstring baseDir(g_cache_dirname.begin(), g_cache_dirname.end());
+    std::wstring searchPattern = baseDir + L"\\*";
+
+    HANDLE hFind = FindFirstFile(searchPattern.c_str(), &data);
+
+    if (hFind != INVALID_HANDLE_VALUE)
+    {
+        do {
+            std::wstring fileName(data.cFileName);
+            if (fileName.compare(L".") != 0 && fileName.compare(L"..") != 0)
```
LarryOsterman: Why use .compare() instead of "=="?

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+            {
+                std::wstring fullFileName = baseDir + L"\\" + fileName;
+
+                throw_if(!DeleteFileW(fullFileName.c_str()),
```
LarryOsterman: What if another instance of the cache is accessing this?

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+    HANDLE file;
+    int i = 0;
+    do {
+        file = CreateFile(lpFileName, dwDesiredAccess, dwShareMode, lpSecurityAttributes,
+            dwCreationDisposition, dwFlagsAndAttributes, hTemplateFile);
+        
+        Sleep(SLEEP_RETRY_MS);
+        i++;
+    } while ((file == INVALID_HANDLE_VALUE) &&
+        (GetLastError() == 32) &&
```
LarryOsterman: Don't use the manifest constant of 32, instead define ERROR_SHARING_VIOLATION as 32 and use that.

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+        (i < MAX_RETRY));
+
+    return file;
+}
+
+
+void local_cache_add(const std::string& id, time_t expiry, size_t data_size, const void* data)
```
LarryOsterman: Does it make sense to have SAL annotations for these parameters? And why is the "data" parameter a "void *" and not a "byte *" (or "uint8_t *").

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+    throw_if(data_size == 0, "Data cannot be empty.");
+    throw_if(data == nullptr, "Data pointer must not be NULL.");
+
+    init();
+    CacheEntryHeaderV1 header{};
+    header.version = CACHE_V1;
+    header.expiry = expiry;
+
+    std::string filename = get_file_name(id);
+    std::wstring wfilename(filename.begin(), filename.end());
```
LarryOsterman: Does this correctly do the ANSI->Unicode conversion?

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+
+    HANDLE file = OpenHandle(wfilename.c_str(), GENERIC_WRITE, 0, NULL, CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, NULL);
+    throw_if(file == INVALID_HANDLE_VALUE, "Open handle failed");
```
LarryOsterman: You've lost the error code here. Also, if you're using a "error propagation is throw" paradigm, then OpenHandle should throw the exception, not this function.

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+
+    OVERLAPPED overlapvar = { 0 };
+
+    throw_if(!LockFileEx(file, LOCKFILE_EXCLUSIVE_LOCK, 0, MAXDWORD, MAXDWORD, &overlapvar),
```
LarryOsterman: You have the file opened for exclusive access - nobody else can have a handle to the file. Why are you bothering to lock the file's contents?

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+    throw_if(!WriteFile(file, &header, sizeof(header), &headerwritten, NULL),
+        "Header write to local cache failed");
+
+    throw_if(!WriteFile(file, data, (DWORD)data_size, &datawritten, NULL),
+        "Data write to local cache failed");
+
+    throw_if(!UnlockFileEx(file, 0, MAXDWORD, MAXDWORD, &overlapvar),
+        "Exclusive write unlock operation failed");
+
+    CloseHandle(file);
```
LarryOsterman: You leak the file handle if any of the "throw_if" calls above throws. That's probably not what you want.

LarryOsterman: In general, when using a "throw" pattern, you need to use RAII techniques to manage resource lifetime (std::unique_ptr, auto-handle classes, etc).

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
 {
-    (void)id;
-    return nullptr;
+    throw_if(id.empty(), "The 'id' parameter must not be empty.");
+    init();
+
+    std::string filename = get_file_name(id);
+    std::wstring wfilename(filename.begin(), filename.end());
+    
+    HANDLE file = OpenHandle(wfilename.c_str(), GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL);
```
LarryOsterman: Use "nullptr" instead of "NULL" throughout the file - nullptr is a built-in type, NULL is a #define to "0", and isn't type safe.

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+    if (file == INVALID_HANDLE_VALUE)
+    {
+        return nullptr;
+    }
+
+    OVERLAPPED overlapvar = { 0 };
+
+    throw_if(!LockFileEx(file, 0, 0, MAXDWORD, MAXDWORD, &overlapvar),
```
LarryOsterman: Why are you locking the file here? What are you trying to protect?

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+    std::string filename = get_file_name(id);
+    std::wstring wfilename(filename.begin(), filename.end());
+    
+    HANDLE file = OpenHandle(wfilename.c_str(), GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL);
+    if (file == INVALID_HANDLE_VALUE)
+    {
+        return nullptr;
+    }
+
+    OVERLAPPED overlapvar = { 0 };
```
LarryOsterman: Since the file wasn't opened for async I/O, why do you have an OVERLAPPED here?

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+        return nullptr;
+    }
+
+    DWORD size = GetFileSize(file, NULL);
+    int datasize = size - sizeof(CacheEntryHeaderV1);
+    auto cache_entry = std::make_unique<std::vector<uint8_t>>(datasize);
+
+    LPDWORD dataread = 0;
+    
+    throw_if(!ReadFile(file, cache_entry->data(), size, dataread, NULL),
```
LarryOsterman: What happens if the read returns less than datasize?

LarryOsterman: ReadFile returns the number of bytes read from the file, but the file might be 0 bytes long. You need to check GetLastError if ReadFile returns 0.

src/UnitTests/test_local_cache.cpp
```diff
@@ -73,7 +78,7 @@ static void VerifyExpiryWorks()
 {
     TEST_START();
 
-    int expiry_seconds = 2;
+    int expiry_seconds = 5;
```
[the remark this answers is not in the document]
msft-chow: Temporary change because of debugging. Reverted

src/UnitTests/test_local_cache.cpp
```diff
@@ -146,14 +157,14 @@ static void ThreadSafetyTest()
...
-    auto cache_writer = [&](int id) {
+    auto cache_writer = [&](void) {
         for (unsigned i = 0; i < THREAD_LOOP_COUNT; ++i)
         {
             local_cache_add(ID, now() + 60, data.size(), data.data());
         }
     };
 
-    auto cache_reader = [&](int id) {
+    auto cache_reader = [&](void) {
```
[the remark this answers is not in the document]
msft-chow: Compiler was complaining about unused variables when there was a id parameter

src/Windows/UnitTest/dcap_provider_ut.vcxproj
```diff
@@ -79,21 +79,25 @@
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
     <ClCompile>
       <Optimization>Disabled</Optimization>
-      <PreprocessorDefinitions>_DEBUG;_LIB;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>_DEBUG;_LIB;_CRT_SECURE_NO_WARNINGS;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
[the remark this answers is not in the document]
msft-chow: Getenv and mbstowcs are depreciated and compiler wants to use other alternatives.

src/UnitTests/test_local_cache.cpp
```diff
@@ -165,16 +176,21 @@ static void ThreadSafetyTest()
     // prime the cache entry first (in case a read thread runs first)
     local_cache_add(ID, now() + 60, data.size(), data.data());
 
+#if defined(__LINUX__)
```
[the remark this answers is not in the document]
msft-chow: Compiler says that std::array<std::thread, 8> is a incomplete type

src/UnitTests/test_local_cache.cpp
```diff
@@ -165,16 +176,21 @@ static void ThreadSafetyTest()
     // prime the cache entry first (in case a read thread runs first)
     local_cache_add(ID, now() + 60, data.size(), data.data());
 
+#if defined(__LINUX__)
     std::array<std::thread, 8> threads;
     for (size_t i = 0; i < threads.size(); ++i)
+#else
+    std::thread threads[8];
+    for (size_t i = 0; i < sizeof(threads) / sizeof(*threads); ++i)
```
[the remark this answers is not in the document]
msft-chow: done

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+        throw std::runtime_error(error);
+    }
+}
+
+static void throw_errno(const std::string& description, int err)
 {
-    // noop
+
+    _locale_t locale = _create_locale(LC_ALL, "POSIX");
+    if (locale == NULL_LOCALE)
```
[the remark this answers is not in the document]
msft-chow: This is from the Linux portion. Throw_errno is called during the make directory process. Locale may not be necessary. I'm not sure

src/Windows/dll/local_cache.cpp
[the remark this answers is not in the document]
msft-chow: changed variable, I will probably update this later to use fileapi

src/UnitTests/test_local_cache.cpp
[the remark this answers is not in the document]
LarryOsterman: Why doesn't that apply to the linux version of this function too?

src/Windows/UnitTest/dcap_provider_ut.vcxproj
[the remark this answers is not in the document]
LarryOsterman: Why not use the non-deprecated alternatives then? In general, disabling security warnings is NOT a good thing.

src/UnitTests/test_local_cache.cpp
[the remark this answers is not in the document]
msft-chow: In the original code, it was initially not being used because it was using ID originally

src/Windows/dll/local_cache.cpp
msft-chow: This was definied in sample code online that I looked at, do you have another way you would suggest?

src/UnitTests/test_local_cache.cpp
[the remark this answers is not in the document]
msft-chow: done

src/Windows/dll/local_cache.cpp
msft-chow: done

src/Windows/dll/local_cache.cpp
msft-chow: This was definied in sample code online that I looked at, do you have another way you would suggest?

src/Windows/UnitTest/dcap_provider_ut.vcxproj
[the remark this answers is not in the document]
msft-chow: Currently rewriting and getting rid of deprecated functions

src/Windows/dll/local_cache.cpp
msft-chow: I was not aware that APPDATA was not saved between sessions, I will be using localappdata

src/Windows/dll/local_cache.cpp
msft-chow: Changed

src/Windows/dll/local_cache.cpp
msft-chow: Done

src/Windows/dll/local_cache.cpp
msft-chow: Is there a difference? I was under the impression that its arbitrary

src/Windows/dll/local_cache.cpp
msft-chow: Done

src/Windows/dll/local_cache.cpp
msft-chow: Changed

src/Windows/dll/local_cache.cpp
msft-chow: In the linux local cache, data is a void*. The data that is being passed in is a uint8 vector but I kept void* to keep in parity with linux. I dont believe SAL annotations are neccessary for the parameters, i believe they are pretty self explanatory

src/Windows/dll/local_cache.cpp
msft-chow: I completely forgot that createfile has a built in exclusive lock. I have removed it and everything still works fine

src/Windows/dll/local_cache.cpp
msft-chow: Should have either a shared lock from read or exclusive lock from write. It will delete after those handles are closed i believe

src/Windows/dll/local_cache.cpp
msft-chow: This was done intentionally to keep in parity with linux version. In the linux version when opening a handle in the add function, if that fails, it should throw back a error but whenever this is done in the get function, it is only supposed to return a nullptr. I am not sure how else to emulate that type of logic

src/Windows/dll/local_cache.cpp
msft-chow: Removed.

src/Windows/dll/local_cache.cpp
msft-chow: Added in a separate check for this.

src/Windows/dll/local_cache.cpp
msft-chow: I added another throw_if that closes the handles before throwing error

src/UnitTests/test_local_cache.cpp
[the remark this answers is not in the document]
LarryOsterman: I'm still seeing this difference, did you push up your code review fixes?

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,412 @@
...
+#include <filesystem>
 
-#include "local_cache.h"
+#define NT_SUCCESS(Status)          (((NTSTATUS)(Status)) >= 0)
 
-//
-// The Windows implementation of the local cache is not yet complete.
-// All methods noop, so we'll never cache any data locally.
-//
+#define STATUS_UNSUCCESSFUL         ((NTSTATUS)0xC0000001L)
```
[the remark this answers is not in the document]
LarryOsterman: nvm, bcrypt returns NT status codes.

src/Windows/dll/local_cache.cpp
msft-chow: Used vector

src/Windows/dll/local_cache.cpp
```diff
@@ -1,33 +1,377 @@
...
+{
+	CreateDirectory(dirname.c_str(), NULL);
+	throw_if(GetLastError() == ERROR_PATH_NOT_FOUND && GetLastError() != ERROR_ALREADY_EXISTS, "Path not found");
+}
+
+static void init_callback()
+{
+	const DWORD buffSize = 65535;
+	
+	LPTSTR env_home = new TCHAR[buffSize];
```
LarryOsterman: This leaks the env_home memory.

LarryOsterman: Another example where a std::unique_ptr would help.

msft-chow: Done

src/UnitTests/test_local_cache.cpp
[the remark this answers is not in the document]
msft-chow: pushed again

src/Windows/dll/local_cache.cpp
msft-chow: How it is done in the Linux version

src/Windows/dll/local_cache.cpp
msft-chow: This is how it was done in the Linux version.

PR #82, opened 2019-08-07
Redirect log output to application event viewer in Windows

msft-jonnadul:


src/Windows/dll/evtx_logging.cpp
```diff
@@ -0,0 +1,106 @@
+#include "evtx_logging.h"
+
+#define DCAP_EVTX_KEY "SYSTEM\\CurrentControlSet\\Services\\EventLog\\Application\\dcap_quoteprov"
+#define DCAP_DLL "dcap_quoteprov.dll"
+
+DWORD check_install_event_log_source()
```
LarryOsterman: IIRC the windows DCAP code has access to wil (Windows Interface Library) and that contains RAII wrappers for registry keys and registry operations.

It may make sense to use them here to avoid potential leaks.

LarryOsterman: see wil::unique_hkey.

src/Windows/dll/evtx_logging.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+        if (last_error == ERROR_SUCCESS)
+        {
+            BYTE dll_path[] = DCAP_DLL;
+            const DWORD types_supported = EVENTLOG_ERROR_TYPE |
+                                          EVENTLOG_WARNING_TYPE |
+                                          EVENTLOG_INFORMATION_TYPE;
+
+            last_error = RegSetValueExA(
+                key, "EventMessageFile", 0, REG_SZ, dll_path, sizeof(dll_path));
```
LarryOsterman: The dll_path should probably be an absolute path, not a relative path.

src/Windows/dll/evtx_logging.h
```diff
@@ -0,0 +1,87 @@
+#ifndef _EXAMPLE_EVENT_LOG_MESSAGE_FILE_H_
+#define _EXAMPLE_EVENT_LOG_MESSAGE_FILE_H_
```
LarryOsterman: I don't see the message file updates to the resource file for the DLL to add these message resources, did I miss it?

msft-jonnadul:
> Does the .bin file need to be included in source control? It's the output of mc.exe, so I don't knowwhy it's included here.

Agreed, would be ideal to be able to dynamically generate this .bin file during compile-time, however it seems non-trivial. Will address as part of a future bugfix.

msft-jonnadul:
Whoops accidentally closed.

msft-jonnadul:
> > Does the .bin file need to be included in source control? It's the output of mc.exe, so I don't knowwhy it's included here.
> 
> Agreed, would be ideal to be able to dynamically generate this .bin file during compile-time, however it seems non-trivial. Will address as part of a future bugfix https://github.com/microsoft/Azure-DCAP-Client/issues/83.

PR #91, opened 2020-02-25
Fixes for automated build

deschuma:
1. Fix a build break
2. Workaround an issue with using getenv in release builds which resulted in tests failing

### DIFF
--- a/src/dcap_provider.cpp
+++ b/src/dcap_provider.cpp
@@ -69,12 +69,33 @@ static char PROCESSOR_CRL_NAME[] = "https%3a%2f%2fcertificates.trustedservices.i
 
 static std::string get_env_variable(std::string env_variable)
 {
+#ifdef LINUX
     const char* env_value = getenv(env_variable.c_str());
-
     if (env_value == NULL)
     {
         return std::string();
     }
+#else
+    char* env_value = (char*)malloc(MAX_ENV_VAR_LENGTH);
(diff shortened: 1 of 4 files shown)


src/dcap_provider.cpp
```diff
@@ -69,12 +69,33 @@ static char PROCESSOR_CRL_NAME[] = "https%3a%2f%2fcertificates.trustedservices.i
...
 {
+#ifdef __LINUX__
     const char* env_value = getenv(env_variable.c_str());
-
     if (env_value == NULL)
     {
         return std::string();
     }
+#else
+    char* env_value = (char*)malloc(MAX_ENV_VAR_LENGTH);
```
msft-jonnadul: Use unique_ptr<string> here so you don't have to worry about freeing it later on.

src/logging.cpp
```diff
@@ -1,5 +1,4 @@
 // Licensed under the MIT License.
-#define _CRT_SECURE_NO_WARNINGS
```
msft-jonnadul: Have you checked how this builds in both Linux and Windows?

src/dcap_provider.cpp
deschuma: thanks! I've implemented the change

msft-jonnadul:
LGTM, feel free to merge.

PR #93, opened 2020-03-14
Pass along eppid to the PCK caching service

deschuma:
- Enables the caching service to perform on demand registration by supplying EPPID when getting certs
- Default the collateral version to V1 if it is not specified in the environment variable
- Rename installer and nuget package


src/dcap_provider.cpp
```diff
@@ -118,24 +124,39 @@ static std::string get_env_variable(std::string env_variable)
...
         }
+
         log(SGX_QL_LOG_INFO,
             "Using %s envvar for collateral version URL, set to '%s'.",
             ENV_AZDCAP_COLLATERAL_VER,
             collateral_version.c_str());
-    }
-    return collateral_version;
+        return collateral_version;
+    }
```
yentsanglee: >     [](start = 5, length = 4)

nit: extra spaces

src/dcap_provider.cpp
```diff
@@ -411,7 +435,17 @@ static std::string build_pck_cert_url(const sgx_ql_pck_cert_id_t& pck_cert_id)
     pck_cert_url << '/' << qe_id;
     pck_cert_url << '/' << cpu_svn;
     pck_cert_url << '/' << pce_svn;
-    pck_cert_url << '/' << pce_id << '?';
+    pck_cert_url << '/' << pce_id;
+    if (!eppid.empty())
+    {
+        pck_cert_url << '/' << eppid;
```
yentsanglee: > eppid [](start = 30, length = 6)

This adds 768 chars to the URL. URL size limit is around 2K. Be careful not to exceed that limit.

src/Windows/curl_easy.cpp
```diff
@@ -218,24 +235,51 @@ curl_easy::~curl_easy()
...
+                GetLastError(), "curl_easy::perform/WinHttpSendRequest");
+        }
 
-    //  Wait for the response from the server.
-    if (!WinHttpReceiveResponse(request.get(), nullptr))
-    {
-        throw_on_error(GetLastError(), "curl_easy::perform/WinHttpReceiveRequest");
-    }
+        //  Wait for the response from the server.
+        if (!WinHttpReceiveResponse(request.get(), nullptr))
```
yentsanglee: >WinHttpReceiveResponse [](start = 13, length = 22)

Timeout value is not explicitly set. How long does it time out?

src/Windows/AzureDCAPSetupProject/Product.wxs
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
-  <Product Id="*" Name="Azure DCAP Client" Language="1033" Version="1.0.0.0" Manufacturer="Microsoft" UpgradeCode="8c473d43-cce9-448e-b3ad-3858f4fe6459">
+  <Product Id="*" Name="AzureDCAPClient" Language="1033" Version="1.0.0.0" Manufacturer="Microsoft" UpgradeCode="8c473d43-cce9-448e-b3ad-3858f4fe6459">
```
LarryOsterman: Should probably call it MicrosoftAzureDCAPClient based on feedback from Azure SDK folks.

src/dcap_provider.cpp
LarryOsterman: Alternatively, instead of encoding all this in the URL, possibly put it in the body?

msft-jonnadul:
LGTM, please feel free to merge in once you have perform all appropriate testing.

src/dcap_provider.cpp
yentsanglee: Just as a reference - this is how SGX Attestation Service defined [theirs](https://api.portal.trustedservices.intel.com/documentation#pcs-certificate)

https://api.trustedservices.intel.com/sgx/certification/v2/pckcert?encrypted_ppid={}&cpusvn={}&pcesvn={}&pceid={}" -H "Ocp-Apim-Subscription-Key: {subscription key}

src/Windows/curl_easy.cpp
deschuma: The default timeout is 30 seconds

src/dcap_provider.cpp
deschuma: We can't unify precisely to their API since we can't use the EPPID for identity as far as I know, since it is salted (we use QE ID for this). It's slightly easier to not use name-value pairs to handle cases where a parameter isn't specified. Parsing wise it doesn't matter.

src/dcap_provider.cpp
deschuma: This is a result of running the format rules against the file

src/dcap_provider.cpp
deschuma: I've moved the EPPID into the request body

src/Windows/AzureDCAPSetupProject/Product.wxs
deschuma: Fixed

Issue #95, opened 2020-04-09
Idempotency broken with InstallAzureDCAP.ps1 script & incorrect system path set


ionutbalutoiu:
The following block of code from `InstallAzureDCAP.ps1` script:
https://github.com/microsoft/Azure-DCAP-Client/blob/35f155e1aa4fd0d14dc266e57461816ecd4460e4/src/Windows/GeneratePackage/InstallAzureDCAP.ps1#L5-L19

has three problems:
1. If the destination directory `$localPath` doesn't exist, the dll is not copied unless the script is called again.
2. Idempotency is broken. If the script is called multiple times with the same value for `$localPath`, that is appended to the system path, every time. We should check if the path is already in the system path.
3. The PowerShell `$env:Path` variable is a concatenation of the `User` path and `System` path. Appending to this value and then set it as a system path, can cause `User` paths present in the `System` path environment variable. We should append only to the existing system path.

PR #96, opened 2020-04-09
Fix InstallAzureDCAP.ps1 script

ionutbalutoiu:
* Copy DLL even if destination directory doesn't exist already.
* Fix script idempotency. The `$localPath` is not appended to the
  system path if it already exists.
* Instead of using the `$env:Path`, which is a concatenation of
  the user path and system path, we use the existing value of the
  system path from registry.

Fixes https://github.com/microsoft/Azure-DCAP-Client/issues/95

Signed-off-by: Ionut Balutoiu <ibalutoiu@cloudbasesolutions.com>

### DIFF
--- a/src/Windows/GeneratePackage/InstallAzureDCAP.ps1
+++ b/src/Windows/GeneratePackage/InstallAzureDCAP.ps1
@@ -2,18 +2,45 @@ param (
     [Parameter(Mandatory=$true)][string]$localPath
 )
 
+
+function Add-ToSystemPath
+{
+    Param(
+        [Parameter(Mandatory=$false)]
+        [string[]]$Path
+    )
+    if(!$Path)
+    {
+        return
+    }
+    $registryLocation = 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment'
+    $systemPath = (Get-ItemProperty -Path $registryLocation -Name Path).Path.Split(';')
+    $currentPath = $env:PATH.Split(';')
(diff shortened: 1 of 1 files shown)


deschuma:
I see the build is failing - looks like there was a software configuration change - I'll see if I can track it down

ionutbalutoiu:
> I see the build is failing - looks like there was a software configuration change - I'll see if I can track it down

We've updated our infrastructure Jenkins Azure managed images.

Build looks good now.

deschuma:
The registry changes require the install script to run as admin - I'd prefer not to do that unless we need to. Isn't it the case we can just update the users environment path?

ionutbalutoiu:
> The registry changes require the install script to run as admin - I'd prefer not to do that unless we need to. Isn't it the case we can just update the users environment path?

Well, I just maintained the script logic (it already appended to system PATH).
And I fixed the issues from https://github.com/microsoft/Azure-DCAP-Client/issues/95.

What you mention makes sense to me as well. I don't think we necessarily need to append to system PATH.

I can add another commit and append to user PATH instead of the system PATH, if you want.

deschuma:
I see - I just took a closer look at how OE SDK uses this and what it was doing before - looks good to me please feel free to merge

ionutbalutoiu:
> I see - I just took a closer look at how OE SDK uses this and what it was doing before - looks good to me please feel free to merge

Thank-you for the review! I will merge this.

Issue #88, opened 2019-12-12
Local caching of collaterals in Azure DCAP client


jazzybluesea:
Opening this issue in order to keep track of the offline request.

Currently the collaterals, QE identity info and Revocation info are not cached, only the PCK Cert is cached.  In order to improve performance, caching for these is needed.  For more information please see https://github.com/openenclave/openenclave/issues/2283.

deschuma:
This has been fixed - please feel free to re-open if needed

PR #105, opened 2020-05-06
Cache all client collateral 

deschuma:
Addresses Issue #88 
- Cache CRLs, QE Identity ,and TCB Info for 12 hours 
- Continue caching PCK Certificates for 24 hours

### DIFF
--- a/src/UnitTests/test_quote_prov.cpp
+++ b/src/UnitTests/test_quote_prov.cpp
@@ -12,6 +12,7 @@
 #include <memory>
 #include <sstream>
 #include <sys/stat.h>
+#include <chrono>
 
 #if defined(__LINUX__)
 #include <tgmath.h>
@@ -22,6 +23,8 @@
 #include <windows.h>
 #include <AclAPI.h>
 #endif
+
+using namespace std;
 
 #if defined __LINUX__
 typedef void * libary_type_t;
@@ -34,6 +37,8 @@ typedef struct _access_permission
     ACCESS_MODE access_mode;
 } permission_type_t;
 #endif
+
+typedef void (*measured_function_t)(void);
 
 typedef quote3_error_t (*sgx_ql_get_quote_config_t)(
     const sgx_ql_pck_cert_id_t* p_pck_cert_id,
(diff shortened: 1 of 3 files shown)


src/UnitTests/test_quote_prov.cpp
```diff
@@ -378,16 +378,21 @@ constexpr auto CURL_TOLERANCE = 0.04;
...
+    double duration_curl_cert;
+    double duration_local_cert;
+    double duration_curl_verification;
+    double duration_local_verification;
     local_cache_clear();
 
     start = std::clock();
     GetCertsTest();
-    duration_curl = (std::clock() - start) / (double)CLOCKS_PER_SEC;
+    duration_curl_cert = (std::clock() - start) / (double)CLOCKS_PER_SEC;
```
achamayou, line 389: std::clock isn't a very good approximation for a wall clock (which is presumably what you are looking for here). I would suggest std::chrono::system_clock instead (which produces time_points, from which you can get typed durations etc).

deschuma: Thanks! I cleaned this up a bit and am using std::chrono::system_clock

src/Windows/dll/dcap_provider.def
```diff
@@ -1,12 +1,10 @@
-LIBRARY dcap_quoteprov
-EXPORTS
-    sgx_ql_get_quote_config
-    sgx_ql_free_quote_config
-    sgx_ql_get_revocation_info
-    sgx_ql_free_revocation_info
-    sgx_get_qe_identity_info
-    sgx_free_qe_identity_info
```
jazzybluesea, line 8: Removing sgx_get/free_qe_identity_info will force users to have to use the latest OE sdk.  I suggest keeping this at least for one release to make it easier for users to upgrade one library at a time?

src/dcap_provider.cpp
```diff
@@ -707,21 +769,83 @@ static std::string build_enclave_id_url(
...
+        response_body = curl_operation->get_body();
         auto get_header_operation =
-            get_unescape_header(*crl_operation, header_name, &issuer_chain);
-        return convert_to_intel_error(get_header_operation);
+            get_unescape_header(*curl_operation, header_name, &issuer_chain);
+
+        retval = convert_to_intel_error(get_header_operation);
+
+        // Update the cache if needed
+        time_t expiry = 0;
```
LarryOsterman, line 832: You cache even if there's an error? But the cache doesn't seem to include the error, so I'm not 100% sure why you'd cache on failure here.

deschuma: Thanks! Definitely don't want to be caching on error - fixed

src/Windows/dll/dcap_provider.def
deschuma: I've brought these back in - will remove in next release. Since we aren't testing these anymore, I really don't like keeping them around. When was this removed from OE SDK?

jazzybluesea: The latest one v0.9, so I think in your next release you should be able to remove it.

src/dcap_provider.cpp
```diff
@@ -236,24 +247,75 @@ static inline quote3_error_t fill_qpl_string_buffer(
...
 //
-// determines the maximum age in local cache
+// Determine time cache should invalidate for given collateral
 //
-sgx_plat_error_t get_cache_max_age(const curl_easy&, time_t* max_age)
+bool get_cache_expiration_time(CollateralTypes collateral_type, time_t &expiration_time)
 {
-    if (max_age == nullptr)
+    time_t max_age = 0;
+    tm* max_age_s = localtime(&max_age);
```
yentsanglee, line 255: >tm* max_age_s = localtime(&max_age); [](start = 4, length = 36)

max_age is not initialized to current time.

src/dcap_provider.cpp
```diff
@@ -236,24 +247,75 @@ static inline quote3_error_t fill_qpl_string_buffer(
...
+
+    switch(collateral_type)
     {
-        return SGX_PLAT_ERROR_INVALID_PARAMETER;
+        case CollateralTypes::TcbInfo:
+        case CollateralTypes::QeIdentity:
+        case CollateralTypes::QveIdentity:
+        {
+            max_age_s->tm_hour += 12;
```
yentsanglee, line 263: There is a potential issue - local cache may not be aware an collateral has expired. For example, at 12:00 PM we locally cache a QE Identity what will expire at 1:00 PM on the same day and this QE Identity will stay in local cache for 12 hours. From 1:00 PM to midnight, the attestation will fail due to QE Identity expiration.

deschuma: Since PCK cache will typically update the collateral well before expiration, I think we will be OK here but you are absolutely right I'll fix this

yentsanglee: I am not sure how to fix the issue unless we parse the collaterals?

---
In reply to: [423988419](https://github.com/microsoft/Azure-DCAP-Client/pull/105#discussion_r423988419) [](ancestors = 423988419)

deschuma: Yes you are correct - in the interest of performance and simplicity I think it may be better to rely on the PCK cache update policy. I just confirmed the settings for QE ID and CRLS - they get refreshed 5 days before expiration

Issue #102, opened 2020-04-25
Enable logging if environment variable set


letmaik:
To diagnose issues it would be convenient if the internal printf logging could be enabled if some AZDCAP_LOGGING=1 environment variable or similar is set. Currently, the library has to be rebuilt to enable logging. This is for Linux. I'm aware that on Windows some event logging takes place already, though the same printf approach would make sense for convenience.

letmaik:
I'm afraid not as I don't manage the DCAP library myself but rather use it via Open Enclave. I would consider the proposed environment variable to be something like [LD_DEBUG](http://www.bnikolic.co.uk/blog/linux-ld-debug.html) which works the same way and is very useful without touching the code.

deschuma:
OESDK sets the logging callback - I believe all of that logging is managed via the OE_LOG_LEVEL environment variable already (including the messages from the az dcap library). 

Regardless I think it would be nice to be able to control the logging independently via a separate environment variable but let me know what you think

letmaik:
Ah I didn't realize. So OE turns on DCAP logging if the log level is at least INFO, which doesn't print out too much extra stuff, but still quite a bit. I think I'm still interested in a separate environment variable for two reasons: 1) the DCAP client does network requests and I'm really interested to see them and only them (I'm not interested in general OE logging), and 2) the DCAP client can also be used outside OE.

deschuma:
Thanks! I agree a separate environment variable would be useful for development

deschuma:
I've opened a PR for this - I opted to prefer the logging callback over the stdout printing in the case that the logging callback is set. For OE SDK, it seems the logging callback is only set if OE_LOG_LEVEL is set. This prevent duplicate messages from being printed in the case both environment variables are set. Environment variable is listed in the readme in the PR.

PR #106, opened 2020-05-19
Allow logging outside of setting the logging callback

deschuma:
- Addresses Issue #102 to enable logging to stdout independent of OE SDK
- Implements OE SDK ask to enable testing without on-demand registration, if desired
- Moves the environment variable functionality into a header file


src/environment.h
```diff
@@ -0,0 +1,67 @@
...
+            (strnlen(env_value, MAX_ENV_VAR_LENGTH) == MAX_ENV_VAR_LENGTH))
+        {
+            log(SGX_QL_LOG_ERROR,
+                "Value specified in environment variable %s is either empty or "
+                "expected max length '%d'.",
+                env_variable.c_str());
+            return std::string();
+        }
+
+        auto retval = std::string(env_value);
```
LarryOsterman, line 62: Why save to a local variable here before returning? Why not just `return std::string(env_value);`?

src/environment.h
```diff
@@ -0,0 +1,67 @@
...
+        env_variable.c_str(), env_temp.get(), MAX_ENV_VAR_LENGTH);
+    if (status == 0)
+    {
+        log(SGX_QL_LOG_ERROR,
+            "Failed to retreive environment varible for '%s'",
+            env_variable.c_str());
+        return std::string();
+    }
+#endif
+    else
```
yentsanglee, line 50: >else [](start = 4, length = 4)

This 'else' is unnecessary.  Both 'if' statements exit the function. Visually, it seems odd to share 'else' block by 2 'if' blocks.

src/environment.h
```diff
@@ -0,0 +1,67 @@
...
+    {
+        if ((strnlen(env_value, MAX_ENV_VAR_LENGTH) <= 0) ||
+            (strnlen(env_value, MAX_ENV_VAR_LENGTH) == MAX_ENV_VAR_LENGTH))
+        {
+            log(SGX_QL_LOG_ERROR,
+                "Value specified in environment variable %s is either empty or "
+                "expected max length '%d'.",
```
yentsanglee, line 57: >%d [](start = 38, length = 2)

Missing an integer param for the %d.

soccerGB:
The logging for parameters passed in sgx_ql_get_quote_config() is only on when the target was Not found in the cache. Could you take this chance to enable logging when cache was hit as well?

src/logging.cpp
```diff
@@ -15,36 +15,117 @@
...
 #include <vector>
+#include <mutex>
+#include "environment.h"
 
-#if 0 // Flip this to true for easy local debugging
-static void DefaultLogCallback(sgx_ql_log_level_t level, const char* message)
+using namespace std;
+
+sgx_ql_logging_function_t logger_callback = nullptr;
+static sgx_ql_log_level_t debug_log_level;
```
yentsanglee, line 24: >debug_log_level [](start = 26, length = 15)

Initialize it.

src/logging.cpp
```diff
@@ -15,36 +15,117 @@
...
+// Logging function which doesn't allocate any memory
+//
+void log_message(sgx_ql_log_level_t level, const char* message)
+{
+    init_debug_log();
 
     if (logger_callback != nullptr)
     {
         logger_callback(level, message);
+    }else if (enable_debug_log)
```
yentsanglee, line 123: Linefeed

src/logging.cpp
```diff
@@ -15,36 +15,117 @@
...
+#include <mutex>
+#include "environment.h"
 
-#if 0 // Flip this to true for easy local debugging
-static void DefaultLogCallback(sgx_ql_log_level_t level, const char* message)
+using namespace std;
+
+sgx_ql_logging_function_t logger_callback = nullptr;
+static sgx_ql_log_level_t debug_log_level;
+bool enable_debug_log = false;
```
yentsanglee, line 25: >enable_debug_log [](start = 5, length = 16)

Minor suggestion - enable_debug_log is always coupled with debug_log_level. You can define something like "SGX_QL_LOG_NONE = 0" to replace enable_debug_log being false.

src/environment.h
deschuma: Thanks fixed

src/environment.h
deschuma: Thanks fixed

src/logging.cpp
deschuma: thanks - fixed

src/logging.cpp
deschuma: thanks - fixed

deschuma:
@soccerGB logging for parameters happens regardless of caching or not - is there a specific parameter I am missing?
1: Azure Quote Provider: libdcap_quoteprov.so [INFO]: Fetching quote config from cache: 'https://global.acccache.azure.net/sgx/certificates/v2/e40610aa353e1bb52385aba37e8acadb/0f0f0205ff8007000000000000000000/0a00/0000?clientid=production_client&api-version=2018-10-01-preview'.

PR #107, opened 2020-06-04
Fix crash when no logging selected

deschuma:
This PR fixes a crash which would occur if neither the logging callback nor the azure dcap specific environment variable is set and adds a test for this condition

This was tested against Windows Server 2016 and Ubuntu 16

### DIFF
--- a/src/environment.h
+++ b/src/environment.h
@@ -74,15 +74,4 @@ static std::string get_env_variabe_no_log(
     return std::string(env_value);
 }
 
-static std::string get_env_variable(std::string env_variable)
-{
-    std::string error_message;
-    auto retval = get_env_variabe_no_log(env_variable, error_message);
-    if (!error_message.empty())
-    {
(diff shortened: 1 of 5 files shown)


src/environment.h
```diff
@@ -18,49 +18,60 @@
 
 #define MAX_ENV_VAR_LENGTH 2000
 
-static std::string get_env_variable(std::string env_variable)
+#include <sstream>
+
+static std::string get_env_variabe_no_log(
```
LarryOsterman, line 23: Instead of taking an "out" parameter and returning a string value, would it make sense to return a std::pair (or std::tuple) with two string values - an error and an environment value? That way the calling signature is the same and you can change the error handling pattern to be `auto val = get_env_variable_no_log(xxx); if (!val.second.empty()) { log val.second; }`?

LarryOsterman, line 23: `get_env_variabe_no_log`??? Spelling error?

deschuma: Thanks! fixed

Issue #114, opened 2020-07-15
Error log if AZDCAP_DEBUG_LOG_LEVEL env var is not set


jumaffre:
If the `AZDCAP_DEBUG_LOG_LEVEL` environment is not set, the following error log is printed: 
`Azure Quote Provider: libdcap_quoteprov.so [ERROR]: Could not retreive environment variable for 'AZDCAP_DEBUG_LOG_LEVEL'`

I believe this is undesirable behaviour and the library should set a sensible default value for the log level that _could_ be overridden with `AZDCAP_DEBUG_LOG_LEVEL`.

achamayou:
That should be a warning/info, it's definitely not an error. Filed https://github.com/microsoft/Azure-DCAP-Client/pull/115 for the typo.

PR #122, opened 2020-07-22
Use Cache-Control header to determine how long collateral should be cached for

deschuma:
This PR allows PCK Cache to determine how long to cache collateral. Ordinarily, collateral should always be cached for 24 hours but PCK cache will examine the expiration time of the collateral and notify the Azure DCAP Client how long the collateral should be cached for.  

For example, if the collateral will expire in 3 hours, the cache-control header will tell Azure DCAP Client to cache the collateral for only three hours so Azure DCAP client will re-query for newer collateral in that time frame. 

This PR also addresses issue #114 by re-classifying errors reading environment variables to be warning instead of errors. The library is configured to operate with sensible defaults if no environment variables are defined.

### DIFF
--- a/src/dcap_provider.cpp
+++ b/src/dcap_provider.cpp
@@ -80,6 +80,8 @@ enum class CollateralTypes
     PckRootCrl
 };
 
+using namespace std;
+
 static std::string get_env_variable(std::string env_variable)
 {
     auto retval = get_env_variable_no_log(env_variable);
@@ -203,34 +205,46 @@ static inline quote3_error_t fill_qpl_string_buffer(
 //
 // Determine time cache should invalidate for given collateral
 //
-bool get_cache_expiration_time(CollateralTypes collateral_type, time_t &expiration_time)
+bool get_cache_expiration_time(const string &cache_control, const string &url, time_t &expiration_time)
 {
     time_t max_age = 0;
     tm* max_age_s = localtime(&max_age);
-
-    switch(collateral_type)
+    string match = "max-age=";
+    size_t index = cache_control.find(match);
+    int cache_time_seconds = 0;
+    if (index != string::npos)
     {
-        case CollateralTypes::TcbInfo:
-        case CollateralTypes::QeIdentity:
-        case CollateralTypes::QveIdentity:
+        try 
         {
-            max_age_s->tm_hour += 12;
-            break;
+            cache_time_seconds = stoi(cache_control.substr(index + match.length()));
         }
-        case CollateralTypes::PckCert:
-        case CollateralTypes::PckCrl:
-        case CollateralTypes::PckRootCrl:
+        catch (std::invalid_argument e)
         {
-            max_age_s->tm_mday += 1;
-            break;
+            log(SGX_QL_LOG_ERROR,
+                "Invalid argument thrown when parsing cache-control. Header text: '%s' Error: '%s'",
+                cache_control.c_str(),
+                e.what());
+            cache_time_seconds = 0;
+            return false;
         }
-        default:
+        catch (std::out_of_range e)
         {
+            log(SGX_QL_LOG_ERROR,
+                "Invalid argument thrown when parsing cache-control. Header "
+                "text: '%s' Error: '%s'",
+                cache_control.c_str(),
+                e.what());
+            cache_time_seconds = 0;
             return false;
         }
     }
 
+    max_age_s->tm_sec += cache_time_seconds;
     expiration_time = time(nullptr) + mktime(max_age_s);
+    log(SGX_QL_LOG_INFO,
+        "Caching collateral '%s' for '%d' seconds",
+        url.c_str(),
+        cache_time_seconds);
     return true;
 }
 
@@ -310,7 +324,13 @@ sgx_plat_error_t get_unescape_header(
 
     result = get_raw_header(curl, header_item, &raw_header);
     if (result != SGX_PLAT_ERROR_OK)
+    {
+        log(SGX_QL_LOG_INFO,
+            "Failed to escape header %s\n",
+            header_item.c_str());
         return result;
+    }
+
     *unescape_header = curl.unescape(raw_header);
     log(SGX_QL_LOG_INFO,
         "unescape_header %s:[%s]\n",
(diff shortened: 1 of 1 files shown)


src/dcap_provider.cpp
```diff
@@ -203,34 +205,46 @@ static inline quote3_error_t fill_qpl_string_buffer(
...
-        case CollateralTypes::PckRootCrl:
+        catch (std::invalid_argument e)
         {
-            max_age_s->tm_mday += 1;
-            break;
+            log(SGX_QL_LOG_ERROR,
+                "Invalid argument thrown when parsing cache-control. Header text: '%s' Error: '%s'",
+                cache_control.c_str(),
+                e.what());
+            cache_time_seconds = 0;
```
yentsanglee, line 227: >cache_time_seconds = 0; [](start = 12, length = 23)

cache_time_seconds is local. Setting it before returning has no meaning.

src/dcap_provider.cpp
```diff
@@ -203,34 +205,44 @@ static inline quote3_error_t fill_qpl_string_buffer(
...
 // Determine time cache should invalidate for given collateral
 //
-bool get_cache_expiration_time(CollateralTypes collateral_type, time_t &expiration_time)
+bool get_cache_expiration_time(const string &cache_control, const string &url, time_t &expiration_time)
 {
     time_t max_age = 0;
     tm* max_age_s = localtime(&max_age);
-
-    switch(collateral_type)
+    string match = "max-age=";
```
LarryOsterman, line 212: Why is "match" a local variable? Why not make it either a const static or a string literal with an associated length? Otherwise you're taking a round trip through the allocator on every call for a string constant.

src/dcap_provider.cpp
```diff
@@ -203,34 +205,44 @@ static inline quote3_error_t fill_qpl_string_buffer(
...
+    size_t index = cache_control.find(match);
+    int cache_time_seconds = 0;
+    if (index != string::npos)
     {
-        case CollateralTypes::TcbInfo:
```
LarryOsterman, line 213: What happens if you don't have a cache_control header value? I think it means that going to immediately expire the collateral, is that the intent (and if it is, is that wise?)

src/dcap_provider.cpp
```diff
@@ -310,10 +322,16 @@ sgx_plat_error_t get_unescape_header(
 
     result = get_raw_header(curl, header_item, &raw_header);
     if (result != SGX_PLAT_ERROR_OK)
+    {
+        log(SGX_QL_LOG_INFO,
+            "Failed to escape header %s",
```
LarryOsterman, line 327: "Failed to escape header"? Isn't this function get_unescape_header?

src/dcap_provider.cpp
```diff
@@ -745,7 +763,7 @@ static std::string get_issuer_chain_cache_name(std::string url)
 static quote3_error_t get_collateral(
     CollateralTypes collateral_type,
     std::string url,
-    const char header_name[],
+    const char issuer_chain_header[],
```
LarryOsterman, line 766: Silly question: Why is this a `const char []` and not a `const char *`?

src/dcap_provider.cpp
```diff
@@ -777,19 +795,27 @@ static quote3_error_t get_collateral(
...
         curl_operation->perform();
         response_body = curl_operation->get_body();
-        auto get_header_operation =
-            get_unescape_header(*curl_operation, header_name, &issuer_chain);
+        auto get_issuer_chain_operation =
+            get_unescape_header(*curl_operation, issuer_chain_header, &issuer_chain);
+
+        std::string cache_control;
+        auto get_cache_header_operation =
+            get_unescape_header(*curl_operation, headers::CACHE_CONTROL, &cache_control);
```
LarryOsterman, line 803: Why are you getting the cache control header when retval is not success? Shouldn't it be inside the "if" lower down?

src/dcap_provider.cpp
```diff
@@ -933,10 +964,14 @@ extern "C" quote3_error_t sgx_ql_get_quote_config(
         buf += cert_data_size;
         assert(buf == buf_end);
 
-        time_t expiry;
-        if (get_cache_expiration_time(CollateralTypes::PckCert, expiry))
+        auto retval = convert_to_intel_error(get_cache_header_operation);
+        if (retval == SGX_QL_SUCCESS)
```
LarryOsterman, line 968: If you're going to fail based on the get cache header operation failing, why do all the stuff between 940 and 966? Shouldn't the code which retrieves the cache control header be closer to where it's used?

src/dcap_provider.cpp
deschuma: Thanks! fixed

src/dcap_provider.cpp
deschuma: Fixed - made it a const static and moved with the others at top of the file

src/dcap_provider.cpp
deschuma: That is the intent - idea being if caching service is broken and handing out expired collateral and not populating the cache control header, azdcap won't cache anything so whatever mitigation we apply on the cache service will become active ASAP from the customer perspective without them having to do anything.

Alternative would be to cache for some predefined time.... I'm open to that as well. 24 hours seems like too long - maybe 8 hours?

src/dcap_provider.cpp
deschuma: Thanks fixed

src/dcap_provider.cpp
deschuma: I don't have a preference either way - changed to const char *

src/dcap_provider.cpp
deschuma: Thanks - fixed

src/dcap_provider.cpp
deschuma: Thanks - fixed

src/dcap_provider.cpp
LarryOsterman: I'm ok with any behavior, but it should be explicit.

deschuma: OK I made this explicit in the log message - "Collateral will not be cached"

PR #126, opened 2020-09-14
Create unit test framework, release test pipeline and internal build flow

wewang3:
1. Build a Google unit test framework for both Windows and Linux
2. Build a Linux CDPx pipeline by automatic build, package and sign process (Microsoft internal use)
3. Build a Windows CloudTest flow (Microsoft internal use)
3. Build a Windows CDPx pipeline by automatic build ,package, sign and CloudTest process (Microsoft internal use)
4. Create a Jenkins test pipeline to test after releasing to Linux Software Repository
5. Updated Jenkins pipeline for added Google unit tests.


src/UnitTest/test_local_cache.cpp
```diff
@@ -1,215 +1,254 @@
...
-    InvalidParams();
-    ThreadSafetyTest();
-}
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+
+#include <gtest/gtest.h>
+
+#undef NDEBUG // ensure that asserts are never compiled out
```
deschuma, line 7: Since we are using gtest now, can this be removed and you replace all the C++ asserts with the test framework one?

src/UnitTest/test_quote_prov.cpp
```diff
@@ -1,709 +1,876 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
```
deschuma, line 1: Do you know why this is showing up as a bulk delete then add? It makes it really hard to review... is it a line ending change or something? Can you change it back so we can review just the lines which have changed?

src/Windows/dcap_provider.sln
```diff
@@ -7,32 +7,38 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "dcap_provider", "dll\dcap_p
...
 EndProject
-Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "dcap_provider_ut", "UnitTest\dcap_provider_ut.vcxproj", "{C356FCD3-13E4-40F1-9FDD-D2E68FA97885}"
-EndProject
-Project("{930C7802-8A8C-48F9-8165-68863BCCD9DD}") = "AzureDCAPSetupProject", "AzureDCAPSetupProject\AzureDCAPSetupProject.wixproj", "{A5E8BFB6-3DBE-498F-9431-544D7C09B1F0}"
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "UnitTests", "UnitTests\UnitTests.vcxproj", "{102D922A-E4D5-4B3F-9F83-BB001ACEFCBF}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|x64 = Debug|x64
+		Debug|x86 = Debug|x86
```
deschuma, line 15: Can you remove the x86 configuration? We don't build and deploy x86 do we?

deschuma (changes requested):

.jenkins/JenkinsfileTestLinuxRelease
```diff
@@ -0,0 +1,132 @@
...
+            timeout(GLOBAL_TIMEOUT) {
+			
+                cleanWs()
+                checkout scm
+				
+                def dcap = load pwd() + "/.jenkins/src/Dcap.groovy"
+
+                // Clone the Open Enclave repo
+                dir("${WORKSPACE}/openenclave") {
+                    git url: 'https://github.com/Microsoft/openenclave.git'
```
szutshi-msft, line 55: Nit. This github string and the string at line 51 "/.jenkins/src/Dcap.groovy" are reused a fair bit. Suggestion to move them to constants.

src/UnitTest/test_local_cache.cpp
```diff
@@ -1,215 +1,254 @@
...
+
+    // ensure the data is there
+    assert(nullptr != local_cache_get(__FUNCTION__));
+
+    // wait for expiry, after which the data should be gone
+
+#if defined(__LINUX__)
+    sleep(expiry_seconds);
+#else
+    Sleep(expiry_seconds * 1000);
```
szutshi-msft, line 109: I assume this is for milliseconds? i.e all other platforms supported other than LINUX use milliseconds and Linux uses seconds?

src/UnitTest/test_local_cache.cpp
```diff
@@ -1,215 +1,254 @@
...
+        for (unsigned i = 0; i < THREAD_LOOP_COUNT; ++i)
+        {
+            auto retrieved = local_cache_get(ID);
+            assert(retrieved != nullptr);
+            assert(*retrieved == data);
+        }
+    };
+
+    // prime the cache entry first (in case a read thread runs first)
+    local_cache_add(ID, now() + 60, data.size(), data.data());
```
szutshi-msft, line 194: Nit: Re-use of magic number 60. Suggestion to define it as a local and for readability.

src/UnitTest/test_local_cache.cpp
wewang3: Yes, that can be removed.

src/UnitTest/test_local_cache.cpp
wewang3: That comes from legacy code. I think Windows is milliseconds and Linux is seconds.

src/Windows/dcap_provider.sln
wewang3: Yes, I will remove it.

.jenkins/JenkinsfileTestLinuxRelease
wewang3: This is not a constant. What it do is to find the path of the current dir. In different agents the paths may be different. It cannot be changed as constant.

szutshi-msft: The hardcoded strings can be different in different agents? I'm referring to the hardcoded strings that are re-used.  My suggestion was making those local vars or something re-usable so you didn't have to define it 3+ times. Anyway just a suggestion.

deschuma (approved):

PR #134, opened 2021-03-19
fixed errors in ubuntu20.04

wewang3:
This PR is to fix compile error in ubuntu20.04

### DIFF
--- a/src/Linux/README.MD
+++ b/src/Linux/README.MD
@@ -73,6 +73,15 @@ build-essential package (`sudo apt install build-essential`) and dh-virtualenv
 
 The output deb file is suitable for installation on any amd64 Ubuntu system.
 
+In Ubuntu 20.04, please try this command to install dh-virtualenv:
+sudo apt-get install software-properties-common
+sudo apt-add-repository universe
+sudo apt-get update
+sudo apt-get install virtualenv
+
+You may need this to install fakeroot:
+sudo apt install build-essential devscripts debhelper check libsubunit-dev fakeroot
+
 ## Troubleshooting
 
 1. You're seeing cryptic errors such as ": No such file or directory":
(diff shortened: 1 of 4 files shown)


src/dcap_provider.cpp
```diff
@@ -234,15 +234,15 @@ bool get_cache_expiration_time(const string &cache_control, const string &url, t
                 return false;
             }
         }
-        catch (std::invalid_argument e)
+        catch (std::invalid_argument& e)
```
olkroshk, line 237: [nit]:  `std::invalid_argument&` -> `const std::invalid_argument&`

wewang3: changed to const reference. It is a better practice

src/Linux/README.MD
```diff
@@ -73,6 +73,19 @@ build-essential package (`sudo apt install build-essential`) and dh-virtualenv
 
 The output deb file is suitable for installation on any amd64 Ubuntu system.
 
+In Ubuntu 20.04, please try this command to install dh-virtualenv:
```
deschuma, line 76: Why are these required? (dh-virtualenv and fakeroot)?
dh-virtualenv looks like it is for python
Can fakeroot be avoided if the user has root privlege?

I'm not a familiar with either of these tools, so please let me know if there are new requirements specific to Ubuntu 20.04 which are requiring them.

src/Linux/debian/changelog
```diff
@@ -1,3 +1,8 @@
+az-dcap-client (1.8) stable; urgency=medium
+  * Enable Ubuntu 20.04 package and Fixed compile errors
```
deschuma, line 2: Nit: spelling: comile->compile

deschuma (changes requested):
Code looks good. Just left a few comments about the readme and release notes

deschuma (approved):

olkroshk (approved):
lgtm, code part looks good to me, I don't have a context on the documentation.

PR #139, opened 2021-04-23
Dev/shivr/ice lake

Shivr-ctrl:


src/UnitTest/test_quote_prov.cpp
```diff
@@ -965,11 +942,11 @@ void SetupEnvironment(std::string version)
...
         if (version == "v3")
         {
-            setenv(
-                "AZDCAP_BASE_CERT_URL",
-                "https://americas.test.acccache.azure.net/sgx/certificates",
-                1);
-        }
+			setenv(
+				"AZDCAP_BASE_CERT_URL",
+                "https://dev.test.acccache.azure.net/sgx/certificates",
```
szutshi-msft, line 947: We may not need this entire if block if we're no longer needing to test against dev clusters and instead can run the tests against prod itself. The v3 block was put in because prod THIM didn't have the prod icelake enabled as intel didn't have it enabled. Suggestion to just default to prod in general again like I believe it used to be.

src/UnitTest/test_quote_prov.cpp
```diff
@@ -983,12 +960,11 @@ void SetupEnvironment(std::string version)
         "https://global.acccache.azure.net/sgx/certificates"));
     EXPECT_TRUE(
         SetEnvironmentVariableA("AZDCAP_CLIENT_ID", "AzureDCAPTestsWindows"));
-    // TODO: (ICX) Remove when we move icx to live and prod
     if (!version.empty() && version.compare("v3") == 0)
     {
         EXPECT_TRUE(SetEnvironmentVariableA(
             "AZDCAP_BASE_CERT_URL",
-            "https://americas.test.acccache.azure.net/sgx/certificates"));
+            "https://dev.test.acccache.azure.net/sgx/certificates"));
```
szutshi-msft, line 967: Same here.

src/UnitTest/test_quote_prov.cpp
```diff
@@ -1061,11 +1037,11 @@ TEST(testQuoteProv, quoteProviderTestsV3DataFromService)
     // Get the data from the service
     //
     SetupEnvironment("v3");
-    ASSERT_TRUE(RunQuoteProviderTests());
-    ASSERT_TRUE(RunQuoteProviderTestsICXV3());
-    ASSERT_TRUE(GetQveIdentityTest());
+   // ASSERT_TRUE(RunQuoteProviderTests());
```
szutshi-msft, line 1040: Probably want to uncomment these tests so we have v3 tests running for qve etc.

src/Windows/local_cache.cpp
```diff
@@ -224,12 +224,24 @@ wil::unique_hfile OpenHandle(LPCWSTR lpFileName, DWORD dwDesiredAccess, DWORD dw
...
         file.reset(CreateFile(lpFileName, dwDesiredAccess, dwShareMode, lpSecurityAttributes,
             dwCreationDisposition, dwFlagsAndAttributes, hTemplateFile));
-        Sleep(SLEEP_RETRY_MS);
+
+		retry = false;
+		if (!file && (GetLastError() == ERROR_SHARING_VIOLATION) && (i < MAX_RETRY)) {
+			retry = true;
+		}
+
+		if (retry == true)
```
szutshi-msft, line 238: Nit: I think if(retry) should be sufficient == true is redundant but not a blocking comment.

src/dcap_provider.cpp
```diff
@@ -1466,10 +1464,8 @@ extern "C" quote3_error_t sgx_ql_get_quote_verification_collateral(
 
         if (strcmp(CRL_CA_PLATFORM, pck_ca) == 0)
         {
-            requested_ca = PLATFORM_CRL_NAME;
-
-            // TODO: (ICX) Test and remove this root distinction once we're pulling icx from live
-            root_crl_name = SBX_ROOT_CRL_NAME;
+			requested_ca = PLATFORM_CRL_NAME;
```
szutshi-msft, line 1467: Nit: Please fix indentation here.

src/UnitTest/test_quote_prov.cpp
```diff
@@ -1061,11 +1037,11 @@ TEST(testQuoteProv, quoteProviderTestsV3DataFromService)
...
+   // ASSERT_TRUE(GetQveIdentityTest());
+   ASSERT_TRUE(RunQuoteProviderTestsICXV3());
 
-#if defined __LINUX__
+ #if defined __LINUX__
```
szutshi-msft, line 1044: Nit: Was this space character accidentally added?

src/UnitTest/test_quote_prov.cpp
```diff
@@ -965,11 +942,11 @@ void SetupEnvironment(std::string version)
         setenv("AZDCAP_COLLATERAL_VERSION", version.c_str(), 1);
         if (version == "v3")
         {
-            setenv(
-                "AZDCAP_BASE_CERT_URL",
-                "https://americas.test.acccache.azure.net/sgx/certificates",
-                1);
-        }
+			setenv(
```
szutshi-msft, line 945: Nit: Fix indentation level.

src/UnitTest/test_quote_prov.cpp
```diff
@@ -476,12 +474,10 @@ static void GetCrlTest()
 //
 // Fetches and validates revocation data for SGX
 //
-static void GetCrlTestICXV3()
+static void GetCrlTestICXV3(bool isProduction)
```
szutshi-msft, line 477: if we're not supporting sbx anymore we don't need the flag for isProduction.

szutshi-msft (changes requested):

src/UnitTest/test_quote_prov.cpp
Shivr-ctrl: Removed

src/UnitTest/test_quote_prov.cpp
Shivr-ctrl: Fixed

src/UnitTest/test_quote_prov.cpp
Shivr-ctrl: Added ICVX3 test. Had to change QEID to non sbx part.

src/dcap_provider.cpp
Shivr-ctrl: fixed to make it consistent

src/Windows/local_cache.cpp
Shivr-ctrl: fixed

src/Windows/local_cache.cpp
```diff
@@ -224,12 +224,24 @@ wil::unique_hfile OpenHandle(LPCWSTR lpFileName, DWORD dwDesiredAccess, DWORD dw
     
     wil::unique_hfile file;
     int i = 0;
+    bool retry;
```
olkroshk, line 227: [nit]: lines 226-244 could be reduced with something like this:
```
bool retry = true;
for (int i=0; (i<MAX_RETRY) && retry; ++i) {
    if (i > 0) Sleep(SLEEP_RETRY_MS);
    file.reset(CreateFile(lpFileName, dwDesiredAccess, dwShareMode, lpSecurityAttributes,
        dwCreationDisposition, dwFlagsAndAttributes, hTemplateFile));
    retry = !file && (GetLastError() == ERROR_SHARING_VIOLATION);
}
```

szutshi-msft: My personal Nit preference is to not have if blocks without { } to ensure no user error is made when editing/debugging code.

olkroshk: sure thing, this is just a suggestion, and it is about using the Boolean variable `retry` as it is (i.e. instead of  `if x  is true then foo is true` -> `foo is x` ) + for-loop because there is a check for the upper boundary (`MAX_RETRY`)

szutshi-msft (approved):

src/UnitTest/test_quote_prov.cpp
```diff
@@ -613,35 +610,10 @@ static void GetRootCACrlTest()
...
-// reasons: 1) The windows system timer runs at a 10ms cadence, meaning that
-// you're not going to see 1ms or 2ms intervals. 2) The windows console is
-// synchronous and quite slow relative to the linux console.
-#if defined __LINUX__
+#ifdef __LINUX__
 constexpr auto CURL_TOLERANCE = 0.002;
-constexpr auto CURL_FILESYSTEM_TOLERANCE = 0;
 #else
-constexpr auto CURL_TOLERANCE = 0.04;
-constexpr auto CURL_FILESYSTEM_TOLERANCE = 0.025;
```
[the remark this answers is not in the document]
Shivr-ctrl: There is no need for any additional delay. The delay on windows is due to a bug which is getting fixed in this PR.

PR #140, opened 2021-05-20
Dev/shivr/ice lake

Shivr-ctrl:
1. Default to version V3 for collateral.
2. For V3 collateral return CRL in DER format.
3. Minor fixes:
   - In case environment variable not specified, log message only for INFO debug level.
   - fixed indentation to remove tabs.


src/UnitTest/test_quote_prov.cpp
```diff
@@ -492,9 +501,18 @@ static void GetCrlTestICXV3()
...
 
-    ASSERT_TRUE(collateral->version == 1);
+    ASSERT_TRUE(collateral != nullptr);
+    if (!collateral_version.compare("v1") || !collateral_version.compare("v2"))
+    {
+        ASSERT_TRUE(collateral->version == 1);
+    }
+    else
+    {
+        ASSERT_TRUE(collateral->version == 3);
```
szutshi-msft, line 514: suggestion to make it an explicit check for comparing "v3", as there will be  version 4 etc and we don't want to default to 3 regardless.

src/dcap_provider.cpp
```diff
@@ -1422,6 +1423,22 @@ extern "C" quote3_error_t sgx_ql_get_quote_verification_collateral(
 
     try
     {
+        std::string collateral_version = get_collateral_version();
+        uint32_t version;
+        std::string api_version;
+
+        if (!collateral_version.compare("v1") ||
+            !collateral_version.compare("v2"))
```
szutshi-msft, line 1431: Nit: seeing these more and more everywhere, suggestion to make "v1" and "v2" etc. constants somewhere so there is more consistency and less room for error.

src/UnitTest/test_quote_prov.cpp
Shivr-ctrl: Currently in the code the only values possible are v1/v2/v3 as the version defaults to V3 if no environment variable specified. If an incorrect environment variable is specified - it will default to v3. When we introduce code for v4 - we would have to make explicit check and assign values for v4 at this spot.

src/dcap_provider.cpp
Shivr-ctrl: Will fix

src/UnitTest/test_quote_prov.cpp
szutshi-msft: Just a suggestion, i've always preferred to be explicit because we may not be the ones making the change in the future and it will need to potentially be explicit for "v3" then anyway when there is a "v4" was my thought. Your call.

src/dcap_provider.cpp
```diff
@@ -104,7 +105,7 @@ static std::string get_collateral_version()
 
     if (collateral_version.empty())
     {
-        log(SGX_QL_LOG_WARNING,
+        log(SGX_QL_LOG_INFO,
             "Using default collateral version '%s'.",
             default_collateral_version.c_str());
         return default_collateral_version;
```
yentsanglee, line 111: (Not related to this PR) Line 115 - 117, should it be

```
        if (collateral_version.compare("v1") &&
            collateral_version.compare("v2") &&
            collateral_version.compare("v3"))
```

src/dcap_provider.cpp
```diff
@@ -1422,6 +1423,22 @@ extern "C" quote3_error_t sgx_ql_get_quote_verification_collateral(
...
+        if (!collateral_version.compare("v1") ||
+            !collateral_version.compare("v2"))
+        {
+            version = 1;
+            api_version = API_VERSION_20200212;
+        }
+        else
+        {
+            version = 3;
+            api_version = API_VERSION_20181001;
```
yentsanglee, line 1439: Why is api_version older for v3?

PR #141, opened 2021-05-21
Default collateral version to V3

Shivr-ctrl:
1. Changed the default collateral version to V3.
2. Fixed bug in handling an unknown version number.

Note - Have not changed the quote_collateral version to 3. There is bug in parsing DER in OE. Once that's fixed, will update this version number.


src/dcap_provider.cpp
```diff
@@ -104,16 +104,16 @@ static std::string get_collateral_version()
...
             "Using default collateral version '%s'.",
             default_collateral_version.c_str());
         return default_collateral_version;
     }
     else
     {
-        if (!collateral_version.compare("v1") &&
-            !collateral_version.compare("v2") &&
-            !collateral_version.compare("v3"))
+        if (collateral_version.compare("v1") &&
```
szutshi-msft, line 114: Nit: Wondering if this would make more sense as an else if block and the final else block being the safe return rather than an if within the else block.

szutshi-msft (approved):
Left an overall nit comment. That is my preference, code is cleaner this way as we're mixing a non valid condition with a valid condition. Will approve with that suggestion.

yentsanglee (approved):
:shipit:

PR #148, opened 2021-09-09
Calling into thim agent to get collateral.

GMunjal:
1. Adding call to thim agent for Get cert collateral.
2. Updating the API call to THIM service to use the new noauth API introduced in the service.

Testing:
1. Unit tests pass on Windows.


msft-gumunjal:
Reopening the PR using msft account.

PR #149, opened 2021-09-09
Calling into thim agent to get collateral.

msft-gumunjal:
1. Adding call to thim agent for Get cert collateral.
2. Updating the API call to THIM service to use the new noauth API introduced in the service.

Testing:
1. Unit tests pass on Windows.

### DIFF
--- a/src/dcap_provider.cpp
+++ b/src/dcap_provider.cpp
@@ -54,12 +54,15 @@ static const std::map<std::string, std::string> default_values = {
 // New API version used to request PEM encoded CRLs
 constexpr char API_VERSION_LEGACY[] = "api-version=2018-10-01-preview";
 constexpr char API_VERSION[] = "api-version=2020-02-12-preview";
-
 static char DEFAULT_CERT_URL[] =
-    "https://global.acccache.azure.net/sgx/certificates";
+    "https://global.acccache.azure.net/sgx/certificates/";
 static std::string cert_base_url = DEFAULT_CERT_URL;
 
-static char DEFAULT_CLIENT_ID[] = "production_client";
+static char DEFAULT_THIM_AGENT_URL[] =
+    "http://169.254.169.254/metadata/THIM/sgx/getCerts?";
+static std::string thim_agent_base_url = DEFAULT_THIM_AGENT_URL;
+
+static char DEFAULT_CLIENT_ID[] = "Azure DCAP_client";
 static std::string prod_client_id = DEFAULT_CLIENT_ID;
 
 static char DEFAULT_COLLATERAL_VERSION[] = "v3";
@@ -133,6 +136,25 @@ static std::string get_collateral_version()
             collateral_version.c_str());
         return collateral_version;
     }
+}
+
+static std::string get_agent_base_url()
+{
+    std::string env_base_url = get_env_variable(ENV_AZDCAP_THIM_AGENT_URL);
+
+    if (env_base_url.empty())
+    {
+        log(SGX_QL_LOG_WARNING,
+            "Using default THIM Agent base cert URL '%s'.",
+            thim_agent_base_url.c_str());
+        return thim_agent_base_url;
+    }
+
+    log(SGX_QL_LOG_INFO,
+        "Using %s envvar for base cert URL, set to '%s'.",
+        ENV_AZDCAP_THIM_AGENT_URL,
+        env_base_url.c_str());
+    return env_base_url;
 }
 
 static std::string get_base_url()
@@ -456,10 +478,9 @@ void safe_cast(input_t in, output_t* out)
     *out = static_cast<output_t>(in);
 }
 
-//
-// Build up the URL needed to fetch specific certificate information.
-//
-static std::string build_pck_cert_url(const sgx_ql_pck_cert_id_t& pck_cert_id)
+static void build_pck_cert_url(
+    const sgx_ql_pck_cert_id_t& pck_cert_id,
+    collateral_fetch_url* collateral_url)
 {
     const std::string qe_id =
         format_as_hex_string(pck_cert_id.p_qe3_id, pck_cert_id.qe3_id_size);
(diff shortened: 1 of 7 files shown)


src/dcap_provider.cpp
```diff
@@ -54,12 +54,15 @@ static const std::map<std::string, std::string> default_values = {
 // New API version used to request PEM encoded CRLs
 constexpr char API_VERSION_LEGACY[] = "api-version=2018-10-01-preview";
 constexpr char API_VERSION[] = "api-version=2020-02-12-preview";
-
 static char DEFAULT_CERT_URL[] =
-    "https://global.acccache.azure.net/sgx/certificates";
+    "https://global.acccache.azure.net/sgx/certificates/";
```
sidse-msft, line 58: Once THIM Service deploys new controller to align with Intel, this should be updated. The final URL will be like: 
`https://global.acccache.azure.net/sgx/certification/v3/pckcert?qeid={}&cpusvn={}&pcesvn={}&pceid={}&encrypted_ppid={}`

src/dcap_provider.cpp
```diff
@@ -54,12 +54,15 @@ static const std::map<std::string, std::string> default_values = {
...
 constexpr char API_VERSION[] = "api-version=2020-02-12-preview";
-
 static char DEFAULT_CERT_URL[] =
-    "https://global.acccache.azure.net/sgx/certificates";
+    "https://global.acccache.azure.net/sgx/certificates/";
 static std::string cert_base_url = DEFAULT_CERT_URL;
 
-static char DEFAULT_CLIENT_ID[] = "production_client";
+static char DEFAULT_THIM_AGENT_URL[] =
+    "http://169.254.169.254/metadata/THIM/sgx/getCerts?";
```
sidse-msft, line 62: Once THIM Agent updates its API to be more aligned with Intel, this should be updated. The final URL will be like: `http://169.254.169.254/metadata/THIM/sgx/certification/v3/pckcert?qeid={}&cpusvn={}&pcesvn={}&pceid={}&encrypted_ppid={}`

src/dcap_provider.h
```diff
@@ -6,6 +6,9 @@
 #define PLATFORM_QUOTE_PROVIDER_H
 
 #include <stdint.h>
+#include <sstream>
+
+using namespace std;
```
sidse-msft, line 11: Is this necessary? I thought because you explicitly use `std::stringstream` with `std::` prefix, you don't need to use the namespace?

src/dcap_provider.cpp
```diff
@@ -54,12 +54,15 @@ static const std::map<std::string, std::string> default_values = {
...
+static std::string thim_agent_base_url = DEFAULT_THIM_AGENT_URL;
+
+static char DEFAULT_CLIENT_ID[] = "Azure DCAP_client";
```
sidse-msft, line 65: Why are we changing the client ID? Originally it was `production_client`.

src/dcap_provider.cpp
```diff
@@ -54,12 +54,15 @@ static const std::map<std::string, std::string> default_values = {
 // New API version used to request PEM encoded CRLs
 constexpr char API_VERSION_LEGACY[] = "api-version=2018-10-01-preview";
 constexpr char API_VERSION[] = "api-version=2020-02-12-preview";
```
sidse-msft, line 56:
```suggestion
constexpr char API_VERSION[] = "api-version=2021-07-22-preview";
```

src/dcap_provider.cpp
```diff
@@ -135,6 +138,25 @@ static std::string get_collateral_version()
     }
 }
 
+static std::string get_agent_base_url()
```
sidse-msft, line 141: Is this function necessary? It made sense with THIM cloud service because we could test different cloud endpoints (e.g. Europe vs. Americas). But THIM Agent will always be hosted at the same local IP address.

src/dcap_provider.cpp
```diff
@@ -977,7 +1053,7 @@ extern "C" quote3_error_t sgx_ql_get_quote_config(
             *curl, headers::CACHE_CONTROL, &cache_control);
 
         auto retval = convert_to_intel_error(get_cache_header_operation);
-        if (retval == SGX_QL_SUCCESS)
+        if ((retval == SGX_QL_SUCCESS) and (!thim_agent_success))
```
sidse-msft, line 1056: The word "and" is valid C++ syntax?

sidse-msft (changes requested):
First pass through code. I haven't looked at every file yet; wanted to get some of these resolved before doing another pass.

src/dcap_provider.cpp
msft-gumunjal: Done :)

src/dcap_provider.cpp
msft-gumunjal: I think this will be useful for dev box testing with IMDS and THIM agent running on the dev box.

src/dcap_provider.cpp
msft-gumunjal: I updated this for telemetry purposes. I think it will be easier to summarize the result.

src/dcap_provider.h
msft-gumunjal: Removed :)

src/dcap_provider.cpp
msft-gumunjal: Agreed! I coded this based on current documentation. Will update this

src/dcap_provider.cpp
msft-gumunjal: Agreed! I coded this based on the current prod code. Will update this

src/dcap_provider.cpp
sidse-msft: Gotcha. Be careful with spaces in the URL; otherwise should be fine

src/dcap_provider.cpp
sidse-msft: Scratch that. I think what you might want to do is: 
`constexpr char API_VERSION_THIM_AGENT[] = "api-version=2021-07-22-preview";`

You can use the new API version for the GetCert calls, and the old API version for the other collateral.

src/Linux/curl_easy.cpp
```diff
@@ -92,7 +92,12 @@ std::unique_ptr<curl_easy> curl_easy::create(const std::string& url, const std::
 
     if (p_body != nullptr && !p_body->empty())
     {
-        easy->set_opt_or_throw(CURLOPT_CUSTOMREQUEST, "GET");
+        if (httpVerb == L"POST") {
+            easy->set_opt_or_throw(CURLOPT_POST, 1L);
+        }
+        else {
```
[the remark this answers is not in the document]
msft-gumunjal: I dont think we have a style checker. Let me try to add one.

src/Linux/curl_easy.cpp
```diff
@@ -70,7 +70,7 @@ char const* curl_easy::error::what() const noexcept
 ///////////////////////////////////////////////////////////////////////////////
 // curl_easy implementation
 ///////////////////////////////////////////////////////////////////////////////
-std::unique_ptr<curl_easy> curl_easy::create(const std::string& url, const std::string* const p_body)
+std::unique_ptr<curl_easy> curl_easy::create(const std::string& url, const std::string* const p_body, LPCWSTR httpVerb))
```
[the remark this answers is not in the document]
msft-gumunjal: The main reason for this file is to keep the structure consistent with the other curl_easy.cpp file(in src/Windows). The variable is used by "WinHttpOpenRequest" in src/Windows/curl_easy.cpp file.

src/Linux/curl_easy.cpp
```diff
@@ -92,7 +92,12 @@ std::unique_ptr<curl_easy> curl_easy::create(const std::string& url, const std::
 
     if (p_body != nullptr && !p_body->empty())
     {
-        easy->set_opt_or_throw(CURLOPT_CUSTOMREQUEST, "GET");
+        if (httpVerb == L"POST") {
```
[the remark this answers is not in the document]
msft-gumunjal: Will update

PR #160, opened 2022-06-30
Dummy PR to validate jenkins run.

msft-gumunjal:


msft-gumunjal:
This was a dummy PR to validate test failures in acc tests. The fix to mitigate ACC test errors was to add a command in shell scripts present in this PR. The change in README was to check if the PR merge was failing because of changes to Azure DCAP code or not.

PR #170, opened 2022-11-18
TDX certificate retrieval

FranciscoJavierOrtegaPalacios:
The problem
We need a way to retrieve certificates for TDX.

What was done
Added a new TDX specific build that retrieves the certificates according to the constraints of a TDX scenario.
Added an API to be able to push a certificate into cache for DCAP to read.

How it was tested
Manual testing for TDX changes
OE and Azure DCAP e2e and unit tests were ran to ensure no regression in existing scenarios.


src/Linux/curl_easy.cpp
```diff
@@ -20,6 +22,12 @@
 #include <strsafe.h>
 #endif
 
+///////////////////////////////////////////////////////////////////////////////
+// Constants
+///////////////////////////////////////////////////////////////////////////////
+static constexpr int maximum_retries = 5;
```
szutshi-msft, line 28: Should we capitalize these constants?

src/Linux/curl_easy.cpp
```diff
@@ -136,7 +144,21 @@ curl_easy::~curl_easy()
...
+	bool finished = false;
+	int retry_delay = initial_retry_delay_ms;
+	CURLcode result;
+	
+	for(int i = 0; i < maximum_retries && !finished; i++){
+		result = curl_easy_perform(handle);
+		
+		if(result == CURLE_OPERATION_TIMEDOUT){
+			std::this_thread::sleep_for(std::chrono::milliseconds(retry_delay));
+			retry_delay *= 2;
```
szutshi-msft, line 156: Is 20 ms even worthwhile to delay? I.e What use case would resolve in 20 ms or even at the best in this case 320 ms?

src/UnitTest/test_quote_prov.cpp
```diff
@@ -126,7 +126,7 @@ static constexpr uint8_t TDX_TEST_FMSPC[] =
     {0x00, 0x80, 0x6F, 0x05, 0x00, 0x00};
 
 const uint16_t custom_param_length = 45;
-const char *custom_param = "tcbEvaluationDataNumber=11;region=us central";
+const char* custom_param = "tcbEvaluationDataNumber=11;region=us central";
```
szutshi-msft, line 129: Put a comment here for future developers why you're using TCBevalDataNum 11 specifically and why this may not be useful later on.

src/dcap_provider.cpp
```diff
@@ -80,20 +80,25 @@ static char PRIMARY_CERT_URL[] =
     "http://169.254.169.254/metadata/THIM/sgx/certification";
 static std::string primary_cert_url = PRIMARY_CERT_URL;
 
+#if defined __SERVICE_VM__
+static char SECONDARY_CERT_URL[] = "http://localhost:4321/tdx/certification";
```
szutshi-msft, line 84: Should these names be more specific? For example THIM_AGENT_TDX_CERT_URL? and similar for PRIMARY above?

FranciscoJavierOrtegaPalacios: The way we are implementing TDX certificate retrieval is by having a specific build for Service VM, the intention being eliminating any chances of running the SGX path within the service VM (or viceversa) due to an error. This means that we can't have different variables for SGX and TDX.

Primary refers to THIMService url, which isn't used in TDX.

szutshi-msft: Oh ok that makes sense! -- Thanks for clarifying. Would it make sense then to call it THIM_AGENT_CERT_URL and THIM_SERVICE_CERT_URL? -- downstream this feels like it makes more sense than primary/secondary.

src/Linux/curl_easy.cpp
FranciscoJavierOrtegaPalacios: We only do this delay in case of ERROR_WINHTTP_TIMEOUT or ERROR_WINHTTP_RESEND_REQUEST. In these cases we already have waited for a bit or have a confirmation that we want to retry as soon as possible.

src/dcap_provider.cpp
```diff
@@ -80,20 +80,25 @@ static char PRIMARY_CERT_URL[] =
     "http://169.254.169.254/metadata/THIM/sgx/certification";
 static std::string primary_cert_url = PRIMARY_CERT_URL;
```
codingedward, line 81: Don't we need to conditionally compile the primary URL too?

src/dcap_provider.cpp
```diff
@@ -896,6 +901,17 @@ static void pck_cert_url(
     url << API_VERSION_07_2021;
 }
 
+static stringstream build_cache_url(const std::string& qe_id, const std::string& cpu_svn, const std::string& pce_svn, const std::string& pce_id) {
```
codingedward, line 904: Thanks for extracting this, it will make easier to make the cache keys consistent

src/dcap_provider.cpp
FranciscoJavierOrtegaPalacios: Primary url is for service, which we don't use in Service VM scenario.

src/dcap_provider.cpp
FranciscoJavierOrtegaPalacios: Yeah, if we ever have the time to refactor DCAP we should do it.

msft-gumunjal: Thim agent is an internal term, so primary and secondary makes more sense. It can be updated to anything.

msft-gumunjal (approved):

src/UnitTest/test_quote_prov.cpp
FranciscoJavierOrtegaPalacios: For this specific test the specific version doesn't matter, just that it can retrieve any specific tcb version with a custom parameter.

src/Linux/curl_easy.cpp
FranciscoJavierOrtegaPalacios: Changed

PR #171, opened 2022-11-25
Adding mariner to the build pipeline

FranciscoJavierOrtegaPalacios:
Added building and E2E tests for Linux Mariner
Added Icelake nodes to the ACC Linux hardware test
Updated the pipeline to not use set-output since it's getting disabled in the future
Changed the windows build VM name since the current one is occupied by a VM that has to be debugged.


codingedward (approved):
LGTM!

PR #183, opened 2023-06-09
Draft PR to trigger the pipeline

FranciscoJavierOrtegaPalacios:


src/Linux/local_cache.cpp
```diff
@@ -97,6 +100,7 @@ class file
 
         if (this->fd != -1)
         {
+			RETRY_ON_EINTR(::chmod(name.c_str(), mode), "chmod");
```
msft-gumunjal, line 103: indentation

src/Linux/local_cache.cpp
```diff
@@ -24,6 +24,9 @@ constexpr locale_t NULL_LOCALE = reinterpret_cast<locale_t>(0);
 static std::string g_cache_dirname = "";
 static std::mutex cache_directory_lock;
 
+static long int last_cache_read_expiry = -1;
```
msft-gumunjal, line 27: why are these needed?

msft-gumunjal (changes requested):

src/Linux/local_cache.cpp
FranciscoJavierOrtegaPalacios: Fixed

src/Linux/local_cache.cpp
```diff
@@ -171,6 +175,7 @@ static void make_dir(const std::string& dirname, mode_t mode)
     {
         if (S_ISDIR(buf.st_mode))
         {
+			chmod(dirname.c_str(), mode);
```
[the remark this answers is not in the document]
FranciscoJavierOrtegaPalacios: Fixed

src/Windows/local_cache.cpp
```diff
@@ -194,6 +194,23 @@ static std::wstring get_file_name(const std::string& id)
...
+std::string get_cached_file_location(const std::string& id){
+	std::wstring fileName = get_file_name(id);
+	//Turning a wstring into a string in this way truncates the value of the char, 
+	//which can have unexpected results when it involves non-ASCII characters
+	//We don't use non-ASCII characters for the default cache path, so that shouldn't be an issue
+	return std::string(fileName.begin(), fileName.end());
+}
+
+std::string get_last_cache_read_expiry(){
+	/*std::string result("Last cache read expiry is unset. This could be because there's been no cache reads yet, it has been read without checking for cache expiry or an error during cache read before cache expiry is checked.");
```
szutshi-msft, line 206: Can we remove this code if its not being used?

src/Linux/local_cache.cpp
```diff
@@ -171,6 +175,7 @@ static void make_dir(const std::string& dirname, mode_t mode)
     {
         if (S_ISDIR(buf.st_mode))
         {
+            chmod(dirname.c_str(), mode);
```
codingedward, line 178: Please make sure to test that this works and does not error out when it is QGSD executing.

src/Linux/local_cache.cpp
```diff
@@ -268,6 +273,18 @@ static std::string get_file_name(const std::string& id)
     return g_cache_dirname + "/" + sha256(id);
 }
 
+std::string get_cached_file_location(const std::string& id){
+	return get_file_name(id);
+}
+
+std::string get_last_cache_read_expiry(){
+	std::string result("Last cache read expiry is unset. This could be because there's been no cache reads yet, it has been read without checking for cache expiry or an error during cache read before cache expiry is checked.");
```
szutshi-msft, line 281: Is this in the right place? With this positioning no matter what this log is going to be sent even if the cache read expiry is set.

szutshi-msft (changes requested):

src/Linux/local_cache.cpp
```diff
@@ -24,6 +24,9 @@ constexpr locale_t NULL_LOCALE = reinterpret_cast<locale_t>(0);
 static std::string g_cache_dirname = "";
 static std::mutex cache_directory_lock;
 
+static long int last_cache_read_expiry = -1;
+static long int last_cache_read_time = -1;
```
codingedward, line 28: I think having these values here for debugging was fine but for production I would recommend exploring all options on how to enable logging on this level. This feels hackish and wil not be clear to future engineers. 

If we end up keeping these we would also have to guarantee thread safety in accessing/writing to them.

src/Linux/local_cache.cpp
```diff
@@ -268,6 +273,18 @@ static std::string get_file_name(const std::string& id)
     return g_cache_dirname + "/" + sha256(id);
 }
 
+std::string get_cached_file_location(const std::string& id){
+	return get_file_name(id);
+}
```
codingedward, line 278: Is there any reason the caller cannot call get_file_name directly?

src/Linux/local_cache.cpp
```diff
@@ -268,6 +273,18 @@ static std::string get_file_name(const std::string& id)
     return g_cache_dirname + "/" + sha256(id);
 }
 
+std::string get_cached_file_location(const std::string& id){
+	return get_file_name(id);
+}
+
+std::string get_last_cache_read_expiry(){
```
codingedward, line 280: Do we still need this for production? This function's signature makes me think that it is going to return the last read expiry while it is returning a log message string. Might be better to return expiry value and then use it when constructing the log message.

src/Linux/local_cache.cpp
```diff
@@ -361,14 +380,18 @@ std::unique_ptr<std::vector<uint8_t>> local_cache_get(
...
-        return nullptr;
+		last_cache_read_expiry = static_cast<long int> (header.expiry);
+		last_cache_read_time = static_cast<long int> (time(nullptr));
+		if(header.expiry <= time(nullptr)){
+			cache_file.close();
+			unlink(file_name.c_str());
+			// Even if unlink fails, we can just return null. Thus, the return
+			// value is intentionally ignored here.
+			return nullptr;
+		}
```
codingedward, line 394: Please consider configuring your dev environment so that formatting and tabbing/spacing is consistent with what is already on the codebase. This conventions make it easy to read code in the future.

src/Windows/local_cache.cpp
```diff
@@ -194,6 +194,23 @@ static std::wstring get_file_name(const std::string& id)
...
+	if(last_cache_read_expiry != -1){
+		result = "Last cache read expiry value is " + std::to_string(last_cache_read_expiry) + " and it was read when time(nullptr) value was " + std::to_string(last_cache_read_time);
+	}
+	return result;*/
```
codingedward, line 210: I still think we don't need this function but if we finally keep it, please remove commented code.

src/dcap_provider.cpp
```diff
@@ -1607,6 +1600,103 @@ bool check_cache(std::string cached_file_name, sgx_ql_config_t** pp_quote_config
...
+        log(SGX_QL_LOG_ERROR, "Unable to add certificate to local cache.");
+    }
+
+	return SGX_QL_SUCCESS;
+}
+
+extern "C" bool store_certificate(const std::string& qe_id, const std::string& cpu_svn, const std::string& pce_svn, const std::string& pce_id, const std::string& response_body)
+{
+	bool result = false;
+	sgx_ql_config_t* pp_quote_config = nullptr;
```
codingedward, line 1688: Don't we need to free `pp_quote_config`? It looks like there's an allocation in store_certificate_internal but that does not get freed anywhere.

src/dcap_provider.cpp
```diff
@@ -1705,9 +1797,11 @@ extern "C" quote3_error_t sgx_ql_get_quote_config(
             else
             {
                 log(SGX_QL_LOG_INFO,
-                    "Trying to fetch response from local cache.");
+                    "Trying to fetch response from local cache in the following direction: %s.", cached_file_name.str().c_str());
```
codingedward, line 1800: Should this be "in the following location"? "direction" is a bit confusing.

src/dcap_provider.cpp
```diff
@@ -1607,6 +1600,103 @@ bool check_cache(std::string cached_file_name, sgx_ql_config_t** pp_quote_config
...
+		{
+			log(SGX_QL_LOG_ERROR, "Unable to retrieve the certificate expiry when writing to local cache.");
+		}
+    }
+    else
+    {
+        log(SGX_QL_LOG_ERROR, "Unable to add certificate to local cache.");
+    }
+
+	return SGX_QL_SUCCESS;
```
codingedward, line 1682: Shouldn't this be returning `retval`?

src/dcap_provider.cpp
```diff
@@ -1607,6 +1600,103 @@ bool check_cache(std::string cached_file_name, sgx_ql_config_t** pp_quote_config
...
+        time_t expiry = 0;
+        if (get_cert_cache_expiration_time(cache_control, cached_file_name, expiry))
+        {
+			local_cache_add(cached_file_name, expiry, buf_size, *pp_quote_config);
+			log(SGX_QL_LOG_INFO, "Stored certificate in cache in the following direction: %s.", get_cached_file_location(cached_file_name).c_str());
+		}
+		else 
+		{
+			log(SGX_QL_LOG_ERROR, "Unable to retrieve the certificate expiry when writing to local cache.");
+		}
```
codingedward, line 1675: Shouldn't this block be setting `retval` to some error code?

src/dcap_provider.cpp
```diff
@@ -1607,6 +1600,103 @@ bool check_cache(std::string cached_file_name, sgx_ql_config_t** pp_quote_config
...
+    }
+    else
+    {
+        log(SGX_QL_LOG_ERROR, "Unable to add certificate to local cache.");
```
codingedward, line 1679: Same here, do we need to set `retval`?

src/Linux/local_cache.cpp
```diff
@@ -97,6 +100,7 @@ class file
 
         if (this->fd != -1)
         {
+            RETRY_ON_EINTR(::chmod(name.c_str(), mode), "chmod");
```
codingedward, line 103: What does `RETRY_ON_EINTR` do?

codingedward (commented):
This is looking good, just a few suggestions and questions.

src/dcap_provider.cpp
```diff
@@ -1607,6 +1600,103 @@ bool check_cache(std::string cached_file_name, sgx_ql_config_t** pp_quote_config
...
+            cache_control.c_str());
+        time_t expiry = 0;
+        if (get_cert_cache_expiration_time(cache_control, cached_file_name, expiry))
+        {
+			local_cache_add(cached_file_name, expiry, buf_size, *pp_quote_config);
+			log(SGX_QL_LOG_INFO, "Stored certificate in cache in the following direction: %s.", get_cached_file_location(cached_file_name).c_str());
+		}
+		else 
+		{
+			log(SGX_QL_LOG_ERROR, "Unable to retrieve the certificate expiry when writing to local cache.");
```
codingedward, line 1674: Shouldn't this block set the `retval` to an error code?

szutshi-msft (approved):

msft-gumunjal (approved):

src/dcap_provider.cpp
```diff
@@ -1607,6 +1600,107 @@ bool check_cache(std::string cached_file_name, sgx_ql_config_t** pp_quote_config
...
+	sgx_ql_config_t* pp_quote_config = nullptr;
+	nlohmann::json json_body = nlohmann::json::parse(response_body);
+
+	std::stringstream cached_file_name = build_cache_url(qe_id, cpu_svn, pce_svn, pce_id);
+
+	if (store_certificate_internal(cached_file_name.str(), json_body, &pp_quote_config) == SGX_QL_SUCCESS) {
+		result = true;
+	}
+
+	delete[] pp_quote_config;
```
codingedward, line 1699: This free is going to be undefined behaviour because allocation was not done for `new sqx_ql_config_t[]`. 

I am not 100% sure but I think this needs to be:

`delete pp_quote_config`  or `delete[] reinterpret_cast<uint8_t*>(pp_quote_config)` 

You could look at how QGSD frees its memory too as a reference.

FranciscoJavierOrtegaPalacios: This is imitating the way memory is freed by sgx_ql_free_quote_config when called by a caller that had previous allocated the memory of pp_quote_config with sgx_ql_get_quote_config, which was already working previously, so I think it should work properly.

src/dcap_provider.cpp
```diff
@@ -1607,6 +1600,107 @@ bool check_cache(std::string cached_file_name, sgx_ql_config_t** pp_quote_config
...
+		}
+    }
+    else
+    {
+        log(SGX_QL_LOG_ERROR, "Unable to add certificate to local cache.");
+    }
+
+	retval = SGX_QL_SUCCESS;
+
+	return retval;
```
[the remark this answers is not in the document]
codingedward: Ah, okay makes sense

PR #185, opened 2023-06-23
Adding time out in Azure dcap in case local agent does not respond.

msft-gumunjal:


src/Windows/curl_easy.h
```diff
@@ -40,7 +40,8 @@ class curl_easy
     static std::unique_ptr<curl_easy> create(
         const std::string& url,
         const std::string* const p_body,
-        unsigned long dwFlags = WINHTTP_FLAG_SECURE);
+        unsigned long dwFlags = WINHTTP_FLAG_SECURE,
+        bool fetchFromLocalAgent = false);
```
FranciscoJavierOrtegaPalacios, line 44: I think that fetchFromLocalAgent should be called fetchingFromLocalAgent instead, to better signal that we're notifying curl that he's fetching from the local agent, rather than telling curl to go fetch from the local agent.

src/Windows/curl_easy.h
```diff
@@ -40,7 +40,8 @@ class curl_easy
     static std::unique_ptr<curl_easy> create(
```
FranciscoJavierOrtegaPalacios, line 40: Please add these changes to src/Linux/curl_easy.h too, otherwise it breaks the build

src/Windows/curl_easy.h
msft-gumunjal: done

src/Windows/curl_easy.h
msft-gumunjal: thanks for pointing it out. Done

src/Windows/curl_easy.cpp
```diff
@@ -217,6 +218,18 @@ std::unique_ptr<curl_easy> curl_easy::create(
...
 
+    if (fetchingFromLocalAgent)
+    {
+        // Setting nResolveTimeout = 0. this is default value. Since we directly
+        // connect to IP, DNS resolution doesnot make sense.
+        // Setting nConnectTimeout = 60000 ms. This is the default value.
+        // Setting nSendTimeout = 150 ms.
+        // Setting nReceiveTimeout = 150 ms.
+        if (!WinHttpSetTimeouts(
+                curl->sessionHandle.get(), 0, 60000, 150, 150))
```
codingedward, line 229: I'm curious why the timeout is different for Windows/Linux

src/Linux/curl_easy.cpp
```diff
@@ -86,7 +86,11 @@ char const* curl_easy::error::what() const noexcept
 ///////////////////////////////////////////////////////////////////////////////
 // curl_easy implementation
 ///////////////////////////////////////////////////////////////////////////////
-std::unique_ptr<curl_easy> curl_easy::create(const std::string& url, const std::string* const p_body, unsigned long dwflag)
+std::unique_ptr<curl_easy> curl_easy::create(
+    const std::string& url,
+    const std::string* const p_body,
+    unsigned long dwflag,
+    bool fetchingFromLocalAgent)
```
codingedward, line 93: NOTE: Updating the code based the following feedback is optional.

It is usually better to separate low-level from high-level implementation details. This makes low-level code more reusable and context agnostic. In this case, `fetchingFromLocalAgent` is a high-level implementation detail but curl_easy may be used in many other contexts for which "fetching from agent" will not make sense. Maybe calling this `shouldTimeout` will keep it general, an even better approach (IMO) would be to make the caller pass the timeout so that this can be reused in any other places where different timeouts apply.

src/Linux/curl_easy.cpp
```diff
@@ -106,6 +110,14 @@ std::unique_ptr<curl_easy> curl_easy::create(const std::string& url, const std::
     easy->set_opt_or_throw(CURLOPT_FAILONERROR, 1L);
     easy->set_opt_or_throw(CURLOPT_SSLVERSION, CURL_SSLVERSION_TLSv1_2);
 
+    // If request is routed to THIM agent for SGX certificate fetch, set
+    // response
+    // timeout to 500 ms
+    if (fetchingFromLocalAgent)
+    {
+        easy->set_opt_or_throw(CURLOPT_SERVER_RESPONSE_TIMEOUT, 0.5L);
```
codingedward, line 118: `0.5L` is of type `long double` which when passed to `curl_easy_setopt` will just get truncated down to 0 because [CURLOPT_SERVER_RESPONSE_TIMEOUT](https://curl.se/libcurl/c/CURLOPT_SERVER_RESPONSE_TIMEOUT.html) expects a `long` while typecasting `double long` to `long` truncates. I think the least timeout you can set using CURLOPT_SERVER_RESPONSE_TIMEOUT is 1s. 




Alternatively, you can use `CURLOPT_TIMEOUT_MS` but this will apply to the whole request as opposed to just the server response. Personally, would recommend setting CURLOPT_SERVER_RESPONSE_TIMEOUT `1s` unless you can guarantee that client will be able to make request within less than `500ms`.

codingedward (changes requested):
This is looking good! Just one tiny thing that would need to change, along with some questions/suggestions.

src/Linux/curl_easy.cpp
msft-gumunjal: Updated to 1 sec. Thank you for catching this. I did not think about implicit typecasting.

src/Linux/curl_easy.cpp
msft-gumunjal: that's a good point but since the scope is limited here, the current implementation should be okay.

codingedward (approved):
Looks good to me!!

PR #197, opened 2024-02-10
src/Linux/local_cache.cpp: add missing `<stdexcept>` include

trofi:
Without the change build fails on `gcc-13` as:

    local_cache.cpp: In function 'void throw_if(bool, const std::string&)':
    local_cache.cpp:40:20: error: 'runtime_error' is not a member of 'std'
       40 |         throw std::runtime_error(error);
          |                    ^~~~~~~~~~~~~
    local_cache.cpp:17:1: note: 'std::runtime_error' is defined in header '<stdexcept>'; did you forget to '#include <stdexcept>'?
       16 | #include <sys/file.h>
      +++ |+#include <stdexcept>
       17 | #include <sys/stat.h>

### DIFF
--- a/src/Linux/local_cache.cpp
+++ b/src/Linux/local_cache.cpp
@@ -6,6 +6,7 @@
 #include <algorithm>
 #include <cstring>
 #include <mutex>
+#include <stdexcept>
 
 #include <fcntl.h>
 #include <ftw.h>


msft-gumunjal:
I will take a look and update this.

pbeza:
@msft-gumunjal any update?

